PR #5, opened 2019-05-06
Add extra parameters stack file

sanderegg:
Adds extra parameters to the resulting stack file used in the portainer


sanderegg:
It is in theory possible to add any number of parameters through that configuration option.
And truth be said, I think there could be a potential conflict with already set parameters. I don't see now what kind of, but if that is the case, we might need to define additional parameters only for selected services. This would make it less generic though...

PR #19, opened 2019-09-03
adminer,minio stacks

sanderegg:
## What do these changes do?

- [x] adds adminer stack
- [x] adds minio stack (deploys for both single and distributed mode)

## Related issue number

links osparc-simcore#575

## Checklist

- [ ] **Runs in the swarm**


services/maintenance/README.md
```diff
@@ -0,0 +1,29 @@
+# portainer stack
```
pcrespov: The portainer stack under maintenance??

services/minio/.env
```diff
@@ -0,0 +1,6 @@
+MINIO_ACCESS_KEY=12345678
```
pcrespov: rename as `.env-defaults` and autogenerate `.env` if it does not exist (as in simcore Makefile)?

services/minio/docker-compose.yaml.j2
```diff
@@ -0,0 +1,65 @@
+version: '3.7'
```
pcrespov: VERY nice!

services/maintenance/README.md
sanderegg: well, I did not touch this part, remember we did this together when copying the maintenance stack out of simcore. Shall I take care of it?

services/minio/.env
sanderegg: actually was wondering why we have it like this?

sanderegg: will be fixed in next PR

PR #20, opened 2019-09-04
Platform deployment

sanderegg:
## What do these changes do?

- [x] define a consistent workflow for deploying full platform (including logs, monitoring, S3, deployer):
    - graylog, minio, deployer all uses a default .env.config file that defines default ENV variables for single machine deployment (e.g. development machine)
    - single development is deployed using __make up__
- [x] updated script for testing in docker-machines
- [x] use Makefile better practices
- [x] various fixes

## Checklist

- [ ] **Runs in the swarm**


services/deployment-agent/.env.config
```diff
@@ -0,0 +1,8 @@
+# defines which configuration file to use
```
pcrespov: Minor: Should we remove the '.' prefix in these files and just call them ``env.config``  (i.e. environment configuration)? 
The files with prefixed with . denote that they are hidden (e.g. when they are files produced temporarily or a hidden configuration that is not that relevant as .gitignore) but this makes no sense here, right? . In the case of `.env` is some legacy that I still do not fully make sense

PR #28, opened 2019-09-24
traefik stack

sanderegg:
## What do these changes do?

- [x] adds traefik 2.0 stack

## Checklist

- [ ] **Runs in the swarm**


services/traefik/docker-compose.yml
```diff
@@ -0,0 +1,64 @@
...
+      - "443:443"
+      # The Web UI (enabled by --api.insecure=true)
+      - "8080:8080"
+    volumes:
+      # So that Traefik can listen to the Docker events
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ./dynamic_conf.yml:/etc/traefik/dynamic_conf.yml
+    configs:
+      - source: traefik_config.yml
+        target: /etc/traefik/traefik.yml
```
oetiker: how about having he static configuration for traefik in the docker compose too ? as per my example ?

services/traefik/docker-compose.yml
```diff
@@ -0,0 +1,64 @@
...
+
+  whoami:
+    # A container that exposes an API to show its IP address
+    image: containous/whoami
+    init: true
+    deploy:
+      labels:
+        - "traefik.enable=false"
+        - "traefik.http.routers.whoami.rule=Host(`devel.io`) && Path(`/whoami`)"
+        # - "traefik.http.routers.whoami.rule=Host(`devel.io`)"
```
oetiker: is the variant with the Path still required ?

services/traefik/dynamic_conf.yml
```diff
@@ -0,0 +1,11 @@
+tls:
+  certificates:
+  - certFile: /run/secrets/domain.crt
+    keyFile: /run/secrets/domain.key
+    stores:
+    - default
+  stores:
```
oetiker: I guess having strict sni on would be fine ... after all supporting old browsers is a security risk in itself and thus not having a defaultCertificate might be actually an advantage

services/traefik/docker-compose.yml
sanderegg: the other stacks are also done similarly so that it's possible to push them through the portainer api.
This way the config files are automatically pushed together with the compose file.

services/traefik/docker-compose.yml
sanderegg: no I will remove it.

services/traefik/dynamic_conf.yml
sanderegg: you're the expert there. so I should remove the defaultCertificate?

services/traefik/Makefile
```diff
@@ -0,0 +1,138 @@
...
+.PHONY: leave
+leave: ## Leaves swarm stopping all services in it
+	-$(DOCKER) swarm leave -f
+
+.PHONY: clean
+clean: ## Cleans unversioned files
+	@git clean -dxf
+
+.PHONY: info
+info: ## expands all variables and relevant info on stack
```
pcrespov: Minor: Looks totally great! I was just wondering whether we should do a fast review to all the makefiles to homogenize them (e.g. name of variables, targets, etc) within osparc-ops repo.

Same applies to configuration files e.g. `.env` vs `.env.config` etc etc

services/traefik/dynamic_conf.yml
sanderegg: done

PR #33, opened 2019-10-14
prometheus in reverse proxy and additional dashboards in monitoring

sanderegg:
## What do these changes do?

- [x] prometheus is available in path /prometheus
- [x] added base dashboards for minio, rabbitMQ
- [x] added prometheus connection to rabbitMQ
- [x] added prometheus documentation for connecting with minio

## Related issue number

ITISFoundation/osparc-simcore#1034

## Checklist

- [ ] **Runs in the swarm**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ MACHINE_IP = $(shell hostname -I | cut -d' ' -f1)
 
 include repo.config
 
-
+SERVICES := adminer deployment-agent graylog maintenance minio monitoring portainer portus traefik
 
 # TARGETS --------------------------------------------------
 .DEFAULT_GOAL := help
@@ -46,8 +46,17 @@ up-local: .install-fqdn certificates/domain.crt certificates/domain.key .create-
 
 .PHONY: up-devel
 up-devel: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
+	echo -n "Ensure osparc-simcore is running the very first time"
 	bash scripts/local-deploy.sh --devel_mode=1
 	@$(MAKE) info-local
+
+.PHONY: down
+down:
+	@for service in $(SERVICES); do \
+		pushd services/$$service; \
+		$(MAKE) down; \
+		popd; \
+	done
 
 .PHONY: leave
 leave: ## leaves the swarm
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ certificates/domain.key:
 	# domain key/crt files must be located in $< and certificates/domain.crt to be used
 	@echo -n "No $< certificate detected, do you wish to create self-signed certificates? [y/N] " && read ans && [ $${ans:-N} = y ]; \
 	$(MAKE) -C certificates create-certificates; \
-	$(MAKE) -C certificates install-root-certificate;	
+	$(MAKE) -C certificates install-root-certificate;
 
 .PHONY: .create-secrets
 .create-secrets:
@@ -42,6 +42,11 @@ certificates/domain.key:
 .PHONY: up-local
 up-local: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
 	bash scripts/local-deploy.sh
+	@$(MAKE) info-local
+
+.PHONY: up-devel
+up-devel: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
+	bash scripts/local-deploy.sh --devel_mode=1
 	@$(MAKE) info-local
 
 .PHONY: leave
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,7 @@ MACHINE_IP = $(shell hostname -I | cut -d' ' -f1)
 
 include repo.config
 
-SERVICES := adminer deployment-agent graylog maintenance minio monitoring portainer portus traefik
-
+SERVICES = $(sort $(wildcard services/*))
 # TARGETS --------------------------------------------------
 .DEFAULT_GOAL := help
 
@@ -46,16 +45,13 @@ up-local: .install-fqdn certificates/domain.crt certificates/domain.key .create-
 
 .PHONY: up-devel
 up-devel: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
-	echo -n "Ensure osparc-simcore is running the very first time"
 	bash scripts/local-deploy.sh --devel_mode=1
 	@$(MAKE) info-local
 
 .PHONY: down
 down:
 	@for service in $(SERVICES); do \
-		pushd services/$$service; \
-		$(MAKE) down; \
-		popd; \
+		pushd $$service; $(MAKE) down; popd; \
 	done
 
 .PHONY: leave
--- a/scripts/local-deploy.sh
+++ b/scripts/local-deploy.sh
@@ -60,7 +60,10 @@ echo -e "\e[1;33mDeploying osparc on ${MACHINE_FQDN}, using credentials $SERVICE
 # -------------------------------- PORTAINER ------------------------------
 echo
 echo -e "\e[1;33mstarting portainer...\e[0m"
-pushd ${repo_basedir}/services/portainer; make up; popd
+pushd ${repo_basedir}/services/portainer
+sed -i "s/PORTAINER_ADMIN_PWD=.*/PORTAINER_ADMIN_PWD=$SERVICES_PASSWORD/" .env
+make up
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ MACHINE_IP = $(shell hostname -I | cut -d' ' -f1)
 
 include repo.config
 
-
+SERVICES := adminer deployment-agent graylog maintenance minio monitoring portainer portus traefik
```
pcrespov: MINOR: add multiline for better versioning

Makefile
```diff
@@ -44,6 +44,20 @@ up-local: .install-fqdn certificates/domain.crt certificates/domain.key .create-
 	bash scripts/local-deploy.sh
 	@$(MAKE) info-local
 
+.PHONY: up-devel
+up-devel: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
+	echo -n "Ensure osparc-simcore is running the very first time"
```
pcrespov: MINOR: Avoid echo and use ``#`` to comment parts as we discussed
MINOR: if every step is commented , you can prefix most of the commands with ``@``

Makefile
pcrespov: -  Why not dynamically deducing the service names using ``docker-compose config --services`` or simply listing all folders under services, e.g. see make's [wildcard function](https://www.gnu.org/software/make/manual/make.html#Wildcard-Function)??

scripts/local-deploy.sh
```diff
@@ -17,6 +17,30 @@ current_git_url=$(git config --get remote.origin.url)
...
+do
+case $i in
+    --devel_mode=*)
+    devel_mode="${i#*=}"
+    shift # past argument=value
+    ;;
+    ##
+    :|*|--help|-h)
+    echo "$usage" >&2
+    exit 1
```
pcrespov: MINOR: I saw this in an example. Always called my attention that ``myprogram --help`` would exit with failure command ..?

scripts/local-deploy.sh
```diff
@@ -32,10 +56,16 @@ cd $repo_basedir;
 echo
 echo -e "\e[1;33mDeploying osparc on ${MACHINE_FQDN}, using credentials $SERVICES_USER:$SERVICES_PASSWORD...\e[0m"
 
+
+# -------------------------------- PORTAINER ------------------------------
 echo
 echo -e "\e[1;33mstarting portainer...\e[0m"
-pushd ${repo_basedir}/services/portainer; make up; popd
+pushd ${repo_basedir}/services/portainer
+sed -i "s/PORTAINER_ADMIN_PWD=.*/PORTAINER_ADMIN_PWD=$SERVICES_PASSWORD/" .env
```
pcrespov: will not work in mac ... define ``sed`` variable and replace with ``gsed`` if OSX (Darwin)

As a side note: I think compatibility with OSX is so far easy to support ... nonetheless, if it becomes a burden we can also drop it

scripts/local-deploy.sh
```diff
@@ -114,16 +146,19 @@ EOF
...
 
+# -------------------------------- MONITORING -------------------------------
 echo
 echo -e "\e[1;33mstarting monitoring...\e[0m"
 # set MACHINE_FQDN
 pushd ${repo_basedir}/services/monitoring
 sed -i "s|GF_SERVER_ROOT_URL=.*|GF_SERVER_ROOT_URL=https://$MACHINE_FQDN/grafana|" grafana/config.monitoring
 sed -i "s|GF_SECURITY_ADMIN_PASSWORD=.*|GF_SECURITY_ADMIN_PASSWORD=$SERVICES_PASSWORD|" grafana/config.monitoring
 sed -i "s|basicAuthPassword:.*|basicAuthPassword: $SERVICES_PASSWORD|" grafana/provisioning/datasources/datasource.yml
+sed -i "s|--web.external-url=.*|--web.external-url=https://$MACHINE_FQDN/prometheus/'|" docker-compose.yml
```
pcrespov: same issue with ``sed`` here (see comment above)

scripts/local-deploy.sh
```diff
@@ -156,33 +191,37 @@ curl -u $SERVICES_USER:$SERVICES_PASSWORD --header "Content-Type: application/js
...
+
+    # -------------------------------- DEPlOYMENT-AGENT -------------------------------
+    echo
+    echo -e "\e[1;33mstarting deployment-agent for simcore...\e[0m"
+    pushd ${repo_basedir}/services/deployment-agent;
+
+    if [[ $current_git_url == git* ]]; then
+        # it is a ssh style link let's get the organisation name and just replace this cause that conf only accepts https git repos
+        current_organisation=$(echo $current_git_url | cut -d":" -f2 | cut -d"/" -f1)
+        sed -i "s|https://github.com/ITISFoundation/osparc-ops.git|https://github.com/$current_organisation/osparc-ops.git|" deployment_config.default.yaml
```
pcrespov: same issue with ``sed`` here (see comment above)

services/monitoring/docker-compose.yml
```diff
@@ -18,7 +18,7 @@ networks:
 services:
 
   prometheus:
-    image: prom/prometheus
+    image: prom/prometheus:latest
```
pcrespov: shouldn't we fix a version and then have a bot that creates a branch, upgrades to latest version each image and run the tests automatically?? 

We should check tools like https://renovatebot.com/

services/traefik/Makefile
```diff
@@ -185,7 +185,8 @@ remove-root-certificate: ## removes the certificate from the host system
 
 .PHONY: .create_secrets
 .create_secrets:
-	@if ! $(DOCKER) secret ls | grep -w domain.crt >/dev/null; then $(DOCKER) secret create domain.crt secrets/domain.crt; fi;
-	@if ! $(DOCKER) secret ls | grep -w domain.key >/dev/null; then $(DOCKER) secret create domain.key secrets/domain.key; fi;
-	@if [ ! -f secrets/rootca.crt ]; then cp secrets/domain.crt secrets/rootca.crt; fi;
-	@if ! $(DOCKER) secret ls | grep -w rootca.crt >/dev/null; then $(DOCKER) secret create rootca.crt secrets/rootca.crt; fi;
\ No newline at end of file
+	@if [[ ! $$(docker secret ls | grep -w domain.crt) ]]; then docker secret create domain.crt secrets/domain.crt; fi;
```
pcrespov: do we really need ``$(DOCKER)``?? if so, why sometimes we use ``docker``
I think we do not ... since we support WSL and not pure windows ... let us do a fast cleanup?! :-)

services/traefik/.env
```diff
@@ -3,4 +3,4 @@ MACHINE_FQDN=osparc.local
 # docker run --rm --entrypoint htpasswd registry:2 -nb user password | sed -e s/\\$/\\$\\$/g
 # echo $(htpasswd -nb user password) | sed -e s/\\$/\\$\\$/g
 TRAEFIK_USER=admin
-TRAEFIK_PASSWORD=$apr1$WOeQr2FL$NAWGz4.pvaV7/YZw.MHJW1
+TRAEFIK_PASSWORD=$apr1$H7Se3sVQ$jXWsAdlwUhxZd54P.AYhX.
```
pcrespov: Why commit this password?? Why not create a script that writes down the right one (e.g. in the make pipelines the .env can be ``sed`` and the password replaced )

Makefile
sanderegg: removed line

Makefile
sanderegg: done

scripts/local-deploy.sh
sanderegg: @ignapas is creating another PR for Mac.

Did not know Mac was a requirement... more than windows? why? ;)

scripts/local-deploy.sh
sanderegg: same comment

scripts/local-deploy.sh
sanderegg: same comment

services/monitoring/docker-compose.yml
sanderegg: created a separate issue for this. As I did not change any behavior here.
I agree we should try that tool.

services/traefik/Makefile
sanderegg: done

services/traefik/.env
sanderegg: this is a default password (which was wrong before). It is actually replaced using sed already when deploying locally.

scripts/local-deploy.sh
ignapas: oh because it's cooler, Windows is cheap...

PR #34, opened 2019-10-15
Mac compatibility

ignapas:
## What do these changes do?
Introducing some changes to make the scripts Mac-compatible.


scripts/local-deploy.sh
```diff
@@ -6,9 +6,11 @@
 
 sed_i()
 {
-	# based on https://unix.stackexchange.com/a/92907
-	sed -e "$1" "$2" > "$2.new"
-	mv -- "$2.new" "$2"
+	if [[ $OSTYPE == "darwin"* ]]; then
+        gsed -i -e "$1" "$2"
```
pcrespov: what about simply defining a ``SED`` variable and  setting it to ``gsed`` as we used to do in the makefile?

scripts/swarm-node-vbox-test-stacks.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/bash
+
+sed_i()
+{
+	if [[ $OSTYPE == "darwin"* ]]; then
+        gsed -i -e "$1" "$2"
+    else
+        sed -i -e "$1" "$2"
+    fi
+}
```
sanderegg: same comment as above

ignapas:
I am not sure about how to test the ``swarm-node-vbox-test-stacks.sh``...

ignapas:
> Please consider trying comment above thx
> 
> Moreover, consider the possibility that the issues in mac comes from the fact that you have [a very old bash installed](https://itnext.io/upgrading-bash-on-macos-7138bd1066ba).
> 
> More-Moreover, the default shell from now on in mac will not be bash anymore but [zsh](https://support.apple.com/en-us/HT208050)

I use zsh... maybe we just don't want to support mac. But as everything but simcore-ops is working well on mac, I asumed this was something to be fixed. If we don't want to support mac, then we just close this PR.
By "copy-pasting every patch", I don't know what you mean.. can you please be more specific or point to the places where you think copy-pasting has been incorrect?

pcrespov:
> By "copy-pasting every patch", I don't know what you mean.. can you please be more specific or point to the places where you think copy-pasting has been incorrect?
> Also the idea about the shell script library seems nice, but I am not sure about the way to proceed.

I mean that I rather collect all these workarounds in a small library to keep compatibility than having to copy&paste every code snippet every time. E.g. we can collect all under ``scripts/osx-compatibility.sh`` with functions/constants and then include and use the throughout all scripts

scripts/local-deploy.sh
```diff
@@ -4,6 +4,9 @@
 #
 #
 
+# Using osx support functions
+source "$( dirname "${BASH_SOURCE[0]}" )/portable.sh"
```
pcrespov: nice

services/portus/Makefile
```diff
@@ -44,7 +44,11 @@ $(if $(wildcard .env), , $(shell cp .env.config .env))
 include .env
 
 # exports
-export MACHINE_IP = $(shell hostname -I | cut -d ' ' -f1)
+# Machine host IP
+$(if $(IS_OSX), \
```
pcrespov: why not calling scripts/portable.sh from here as well? try something like:
```bash
export MACHINE_IP = $(shell $(realpath $(CURDIR)/../../scripts/portable.sh) get_this_ip)
```

services/traefik/Makefile
```diff
@@ -20,7 +20,11 @@ PWD ?= pwd_unknown
 STACK_NAME = $(notdir $(PWD))
 SWARM_HOSTS = $(shell docker node ls --format={{.Hostname}} 2>/dev/null)
 TEMP_COMPOSE=.stack.${STACK_NAME}.yaml
-MACHINE_IP = $(shell hostname -I | cut -d' ' -f1)
+# Machine host IP
+$(if $(IS_OSX), \
```
pcrespov: same  comment as above

services/portus/Makefile
ignapas: done

Issue #26, opened 2019-09-16
auto-deployer does not take the right git commit when generating the stack file


sanderegg:
the auto-deployer for production is set to take the latest commit in osparc-simcore, branch=staging.
This is wrong as it should be checking out the git release tag in branch staging.

This means the stack file is not on the same stage as the docker images!

sanderegg:
use git checkout tags/<mytagname>
deployer needs to know the pattern and if it needs to use tags when checing out the code. it needs to check when tags are changing.

sanderegg:
fixed

PR #35, opened 2019-10-16
Bug26/deployer tag

sanderegg:
## What do these changes do?

Fix bug osparc-ops #26:
- new setting added in the deployer configuration:
  - tags setting. If empty string or missing, same behaviour as usual,
  a regexp means that the deployment-agent will checkout tags that follow the defined pattern only. It will always deploy the **latest** tag that follows that pattern (.e.g. "v[0-9]+.[0-9]+.[0-9]+" will deploy the latest commit with tag such as v0.1.2)
- fixed logging in debug not working
- add SHA in notification message

## Related issue number

osparc-ops #26

## Checklist

- [ ] **Runs in the swarm**


services/deployment-agent/deployment_config.default.yaml
```diff
@@ -47,7 +47,7 @@ main:
             cat .env-devel >> .env && cat .env.platform >> .env &&
             docker-compose -f services/docker-compose.yml -f docker-compose.deploy.yml config > stack.yml
     stack_file: stack.yml # the output file of the command above, or just the file to use
-    excluded_services: [minio]
+    excluded_services: [webclient, minio]
```
pcrespov: MINOR: there is no more webclient, right? In any case, it does not harm to have it... this way you can deploy old versions of simcore

services/deployment-agent/docker/boot.sh
```diff
@@ -35,11 +36,9 @@ fi
 # RUNNING application ----------------------------------------
 if [[ ${SC_BOOT_MODE} == "debug" ]]
```
pcrespov: this is kind of mixed up. There are two things completely orthogonal
 a) log level: this corresponds to the level used by ``logging``
 b) boot mode: this is how the app is boot and is given by ``SC_BOOT_MODE``: 
         -  ``default``: shall be ``simcore-service-deployment-agent --config $APP_CONFIG --loglevel=$LOG_LEVEL``
         - ``debug-ptvsd``: shall boot as ``python3 -m ptsvsd ...`` (note that this mode requires in the entrypoint to ``python3 -m pip install ptvsd`` while starting as root as well.

services/deployment-agent/src/simcore_service_deployment_agent/auto_deploy_task.py
```diff
@@ -255,37 +258,23 @@
...
             log.info("task completed...")
-async def auto_deploy(app: web.Application):
-    log.info("start autodeploy task")
-    subtasks = await _init_deploy(app)
-    # loop forever to detect changes
-    await _deploy(app, subtasks)
-        
 
+def setup(app: web.Application):
+    app.cleanup_ctx.append(persistent_session)
```
pcrespov: very nice

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -74,48 +96,93 @@ class GitRepo:  # pylint: disable=too-many-instance-attributes, too-many-argumen
...
+async def _update_repository(repo: GitRepo):
+    if repo.pull_only_files:
+        await _git_pull_files(repo.directory, repo.paths)
+    else:
+        await _git_pull(repo.directory)
+
+async def _init_repositories(repos: List[GitRepo]) -> Dict:
+    description = {}
     for repo in repos:
         repo.directory = tempfile.TemporaryDirectory().name
```
pcrespov: MINOR: cleanup [tempfile.TemporaryDirectory()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) when done. Can also use a context manager with returned object

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -74,48 +96,93 @@ class GitRepo:  # pylint: disable=too-many-instance-attributes, too-many-argumen
...
+            else f"{repo.repo_id}:{repo.branch}:{sha}"
+    return description
+
+async def _update_repo_using_tags(repo: GitRepo) -> Optional[str]:
+    log.debug("checking %s using tags", repo.repo_id)
+    # check if current tag is the latest and greatest
+    current_tag = await _git_get_current_matching_tag(repo.directory, repo.tags)
+    latest_tag = await _git_get_latest_matching_tag(repo.directory, repo.tags)
+    # there should always be a tag
+    assert latest_tag
```
pcrespov: MINOR: add some message... notice that if AssertError is raised, there is not message shown... this is not as nice as pytest :-)

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -74,48 +96,93 @@ class GitRepo:  # pylint: disable=too-many-instance-attributes, too-many-argumen
...
+    if current_tag == latest_tag:
+        return
```
pcrespov: MINOR: log.debug this decision ... this will help us follow the flow?

services/deployment-agent/src/simcore_service_deployment_agent/auto_deploy_task.py
```diff
@@ -196,18 +179,27 @@
...
-async def _init_deploy(app: web.Application) -> List[SubTask]:
+async def _init_deploy(app: web.Application) -> Tuple[GitUrlWatcher, DockerRegistriesWatcher]:
     try:
+        log.info("initialising...")
+        # get configs
         app[TASK_STATE] = State.STARTING
         app_config = app[APP_CONFIG_KEY]
         app_session = app[TASK_SESSION_NAME]
-        log.info("initialising...")
+        # wait for portianer to be available
```
pcrespov: MINOR: type portAiner  :-)

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -16,53 +17,74 @@
...
+    cmd = f"cd {directory} && git --no-pager describe --tags --always `git rev-list --tags`"
+    all_tags = await run_cmd_line(cmd)
+    if not all_tags:
+        return []
+    list_tags = all_tags.split("\n")
+    reg_object = re.compile(regexp)
+    for tag in list_tags:
+        if reg_object.fullmatch(tag):
+            return tag
+    return ""
```
pcrespov: Why not returning ``None``?
Actually the returned type does not match ``List[str]`` ... do you mean ``Optional[None, str]. Same applies for `` if not all_tags: return [] `` above.

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -16,53 +17,74 @@
...
-    cmd = "cd {directory} && git --no-pager diff --name-only FETCH_HEAD".format(
-        directory=directory)
+    cmd = f"cd {directory} && git --no-pager diff --name-only FETCH_HEAD"
     modified_files = await run_cmd_line(cmd)
     return modified_files
 
-async def _git_get_logs(directory: Path, branch: str) -> str:
-    cmd = "cd {directory} && git log --oneline {branch}..origin/{branch}".format(
-        directory=directory, branch=branch)
+async def _git_get_logs(directory: Path, branch1: str, branch2: str) -> str:
```
pcrespov: I wonder if you could copy this collection of ``_git_..`` helper functions and add them e.g. in ``servicelib``

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
```diff
@@ -74,48 +96,93 @@ class GitRepo:  # pylint: disable=too-many-instance-attributes, too-many-argumen
...
+async def _update_repo_using_branch_head(repo: GitRepo) -> Optional[str]:
+    modified_files = await _git_diff_filenames(repo.directory)
+    if not modified_files:
+        # no modifications
+        return
+    # get the logs
+    logged_changes = await _git_get_logs(repo.directory, repo.branch, repo.branch)
+    log.debug("Changelog:\n%s", logged_changes)
+    await _update_repository(repo)
+    # check if a watched file has changed
```
pcrespov: I wonder if all the "human readable comments" shall be converted into ``log.debug``. It might help tracing the logic workflow

services/deployment-agent/deployment_config.default.yaml
sanderegg: yes currently the production cluster still has the webclient in.

services/deployment-agent/docker/boot.sh
sanderegg: good catch! actually I'm not using loglevel there but in the config file.
cleaned up.

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
sanderegg: well, this is running in a container and I though that in case of issue it might be better that the folder is still around for debugging.

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
sanderegg: done

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
sanderegg: done

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
sanderegg: again good catch. but BTW it needs to be used as Option[str] = Union[None, str].

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
sanderegg: well maybe.. don't know if we need this elsewhere at the moment.

services/deployment-agent/src/simcore_service_deployment_agent/auto_deploy_task.py
sanderegg: done

services/deployment-agent/src/simcore_service_deployment_agent/git_url_watcher.py
pcrespov: ok as soon as this is not constantly creating new folders ,right?

PR #37, opened 2019-10-23
Fix graylog version and add simcore services to monitoring network

sanderegg:
## What do these changes do?

- [x] fix graylog version to 3.1.2-1 as 3.1.2-2 fail to start
- [x] adds webserver/storage to the monitoring network

## Checklist

- [ ] **Runs in the swarm**

### DIFF
--- a/services/deployment-agent/Makefile
+++ b/services/deployment-agent/Makefile
@@ -22,6 +22,17 @@ PUBLIC_NETWORK := $(public_network)
 endif
 export PUBLIC_NETWORK
 
+# Network that includes all services to monitor
+#  - the idea is that it shall connect osparc stack network so that e.g. cadvisor can monitor ALL the stack
+#  - by default it will create an overal attachable network called monitored_network
+ifeq ($(monitored_network),)
+MONITORED_NETWORK = monitored_network
+else
+MONITORED_NETWORK := $(monitored_network)
+endif
+export MONITORED_NETWORK
+
+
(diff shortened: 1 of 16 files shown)


services/deployment-agent/Makefile
```diff
@@ -22,6 +22,17 @@ PUBLIC_NETWORK := $(public_network)
 endif
 export PUBLIC_NETWORK
 
+# Network that includes all services to monitor
+#  - the idea is that it shall connect osparc stack network so that e.g. cadvisor can monitor ALL the stack
```
pcrespov: you mean simcore stack, right?

services/simcore/docker-compose.deploy.yml
```diff
@@ -14,6 +14,7 @@ services:
         - traefik.http.routers.webserver.middlewares=gzip_compress@file
     networks:
       - public
+      - monitored-net
```
pcrespov: let us try to be consistent with names from the beginning. Either we add ``-net`` to all or not. I kind of like the minimalist approach of not but I leave the choice to you :-)

services/deployment-agent/Makefile
sanderegg: yes corrected. copy/paste issue

services/simcore/docker-compose.deploy.yml
sanderegg: renamed

PR #38, opened 2019-10-23
Fix/postgres exporter

pcrespov:
## What do these changes do?

- sync posgress usr/pass in POSTGRES_EXPORTER_DATA_SOURCE_NAME
- create helpers for makefiles under scripts/*.mk

## Checklist

- [x] **Runs in the swarm**


.vscode-template/settings.json
```diff
@@ -0,0 +1,38 @@
...
+  ],
+  "[python]": {
+    "editor.detectIndentation": false,
+    "editor.tabSize": 4
+  },
+  "[makefile]": {
+    "editor.insertSpaces": false
+  },
+  "python.testing.pytestArgs": [
+    "services/web/server"
```
sanderegg: this is probably not going to work for this repo

scripts/info.mk
```diff
@@ -0,0 +1,30 @@
+.PHONY: info info-debug info-swarm
+
+info: ## displays setup information
+	# setup info:
+	@echo ' Detected OS  : $(IS_LINUX)$(IS_OSX)$(IS_WSL)$(IS_WIN)'
```
sanderegg: this implicitely means you have IS_LINUX, IS_OSX, IS_WSL and IS_WIN everywhere. is that the case? or maybe you should include these in the file as well.

.vscode-template/settings.json
pcrespov: i copied the wrong one... thx

PR #44, opened 2019-11-04
Changes arising from staging deployment using let's encrypt

sanderegg:
## What do these changes do?

- add let's encrypt use case

## Checklist

- [ ] **Runs in the swarm**

### DIFF
--- a/services/graylog/Makefile
+++ b/services/graylog/Makefile
@@ -28,6 +28,10 @@ include $(realpath $(CURDIR)/../../scripts/common.mk)
 up: .init .env ${TEMP_COMPOSE}  ## Deploys graylog stack
 	docker stack deploy -c ${TEMP_COMPOSE} ${STACK_NAME}
 
+.PHONY: up-letsencrypt
+up-letsencrypt: .init .env ${TEMP_COMPOSE}-letsencrypt  ## Deploys graylog stack using let's encrypt
+	docker stack deploy -c ${TEMP_COMPOSE}-letsencrypt ${STACK_NAME}
+
 .PHONY: down
 down: ## Removes the stack from the swarm
 	docker stack rm ${STACK_NAME}
@@ -55,4 +59,9 @@ clean: ## Cleans unversioned files
 .PHONY: ${TEMP_COMPOSE}
 ${TEMP_COMPOSE}: docker-compose.yml
 	@docker-compose -f $< config > $@
+	@echo "${STACK_NAME} stack file created for in $@"
+
+.PHONY: ${TEMP_COMPOSE}-letsencrypt
+${TEMP_COMPOSE}-letsencrypt: docker-compose.yml docker-compose.letsencrypt.yml
+	@docker-compose -f $< -f docker-compose.letsencrypt.yml config > $@
 	@echo "${STACK_NAME} stack file created for in $@"
--- a/services/graylog/docker-compose.letsencrypt.yml
+++ b/services/graylog/docker-compose.letsencrypt.yml
@@ -1,0 +1,7 @@
+version: '3.7'
+services:
+  graylog:
+    deploy:
+      labels:
+        - traefik.http.routers.graylog.rule=Host(`${MONITORING_DOMAIN}`) && PathPrefix(`/graylog`)
+        - traefik.http.routers.graylog.tls.certresolver=lehttpchallenge
--- a/services/graylog/docker-compose.yml
+++ b/services/graylog/docker-compose.yml
@@ -11,6 +11,9 @@ services:
       replicas: 1
       restart_policy:
         condition: on-failure
+      placement:
+        constraints:
+          - node.role==manager
     networks:
       default:
         aliases:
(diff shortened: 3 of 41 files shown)


services/graylog/Makefile
```diff
@@ -56,3 +60,8 @@ clean: ## Cleans unversioned files
 ${TEMP_COMPOSE}: docker-compose.yml
 	@docker-compose -f $< config > $@
 	@echo "${STACK_NAME} stack file created for in $@"
+
+.PHONY: ${TEMP_COMPOSE}-letsencrypt
+${TEMP_COMPOSE}-letsencrypt: docker-compose.yml docker-compose.letsencrypt.yml
```
pcrespov: Instead of variable ``TEMP_COMPOSE`` I started to create a fixed value for this output and add it in the gitignore list. It helps to debug if something goes wrong. 

An example is [.stack-simcore-development.yml](https://github.com/ITISFoundation/osparc-simcore/blob/master/Makefile#L133). It is an output of make in osparc-simcore and compresses

services/graylog/docker-compose.yml
```diff
@@ -84,10 +87,14 @@ services:
         - traefik.http.routers.graylog.rule=PathPrefix(`/graylog`)
         - traefik.http.routers.graylog.entrypoints=https
         - traefik.http.routers.graylog.tls=true
+        #  Let's encrypt use case
```
pcrespov: pls justify commented code, otherwise remove.  thx

services/jaeger/Makefile
```diff
@@ -63,6 +67,11 @@ ${TEMP_COMPOSE}: docker-compose.yml
 	@docker-compose -f $< config > $@
 	@echo "${STACK_NAME} stack file created for in $@"
 
+.PHONY: ${TEMP_COMPOSE}-letsencrypt
+${TEMP_COMPOSE}-letsencrypt: docker-compose.yml docker-compose.letsencrypt.yml
```
pcrespov: same as above

services/minio/.env
```diff
@@ -8,3 +8,5 @@ MINIO_NUM_PARTITIONS=1
 # optionally define the host data folder for each partition (must be the same on each node)
 #MINIO_INTERNAL_VOLUME_DATA_PART0=/path/to/partition/0
```
pcrespov: Justfiy commented. e.g. add "Example" in this case

services/graylog/docker-compose.letsencrypt.yml
```diff
@@ -0,0 +1,7 @@
+version: '3.7'
+services:
+  graylog:
+    deploy:
+      labels:
+        - traefik.http.routers.graylog.rule=Host(`${MONITORING_DOMAIN}`) && PathPrefix(`/graylog`)
+        - traefik.http.routers.graylog.tls.certresolver=lehttpchallenge
```
mguidon: wow. thats enough?

services/minio/docker-compose.yaml.j2
```diff
@@ -39,12 +40,16 @@ services:
         - traefik.http.routers.minio.rule=PathPrefix(`/`)
         - traefik.http.routers.minio.entrypoints=storage
         - traefik.http.routers.minio.tls=true
-        - traefik.http.routers.minio.middlewares=gzip_compress@file
+        # Let's encrypt use-case
+        # - traefik.http.routers.minio.tls.certresolver=lehttpchallenge
+        - traefik.http.routers.minio.middlewares=gzip@docker
```
mguidon: Did you ever test performance impact?

services/graylog/docker-compose.yml
sanderegg: removed.. left over

services/minio/.env
sanderegg: hmm you mean the line just above?

services/graylog/Makefile
sanderegg: yes in this case this one converts to .stack.graylog-letsencrypt.yml. I actually prefer using an automatic variable that also scales better when you create other services instead of renaming all the recipes.
does that make sense or I missed the point?

services/graylog/docker-compose.letsencrypt.yml
sanderegg: yeah well, together with the docker-compose.yml file yes. so about 5-6 labels + the initial traefik configuration.

services/minio/docker-compose.yaml.j2
sanderegg: no. but it was also partly enabled in nginx. so I guess not much difference.

services/graylog/Makefile
pcrespov: My bad. I misinterpreted from the partial file the implementation of ``${TEMP_COMPOSE}``. It does exactly what I meant. I got confused by the name ``${TEMP_COMPOSE}`` since I thought it was created randomly everytime. Perhaps ``letsencrypt-${TEMP_COMPOSE}`` so that it has ``yml`` extension?

services/minio/.env
pcrespov: MINOR: I meant that we need to distinguish between commented lines that are forgotten and those that are e.g. added as examples or notes. Please add some keyword as ``NOTE``, ``TODO`` or ``FIXME`` or ``EXAMPLE`` to avoid somebody from deleting it :-)

Issue #40, opened 2019-10-24
graylog consumes multiline logs as multiple entries


pcrespov:
graylog reports each line of a multiline as a separate log entry. Unfortunately these are the most interesting logs e.g. an exception logs a traceback with the call stack in multiple lines

It seems graylog cannot do anything about it and the logs has to be transmitted "as one" by the intermediares:
 - https://community.graylog.org/t/consuming-multiline-docker-logs-via-gelf-udp/972
 - https://community.graylog.org/t/how-to-stream-docker-logs-with-multiline-to-graylog/5881

In our case, it seems our service [logspout](https://github.com/ITISFoundation/osparc-ops/blob/master/services/graylog/docker-compose.yml#L94) in the graylog stack can deal with it provided the [right configuration](https://github.com/gliderlabs/logspout#multiline-logging)

It might also help to introduce a common formatter and filter at the service level

PR #46, opened 2019-11-07
Graylog multilines

Surfict:
## What do these changes do?

Add a multi line support for logs coming from the Simcore Stack, by adding some environment variables telling to Logspout (module used by Graylog to aggregates all the logs coming from the different containers) to use the multi line support when needed.

## Related issue number

#40 

## Checklist

- [x] **Runs in the swarm**
- [x] If you design a new module, add your user to .github/CODEOWNERS


services/graylog/docker-compose.yml
```diff
@@ -102,6 +102,9 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
     environment:
       - ALLOW_TTY=true
+      - MULTILINE_MATCH=first
+      - MULTILINE_PATTERN='^(ERROR|WARNING|INFO|DEBUG|CRITICAL):'
```
pcrespov: Are you sure that the quotes ``'`` around the pattern are needed? I kind of recall that we had to remove them in the demo to make it work! 
Moreover, here since you use a list for the ``environment`` value, each entry is considered a string, i.e. ``"MULTILINE_PATTERN='^(ERROR|WARNING|INFO|DEBUG|CRITICAL):'"`` and when docker-compose splits into key-value, it will assign the quote symbol to the value!

services/simcore/docker-compose.deploy.yml
```diff
@@ -1,3 +1,6 @@
+
+# LOGSPOUT_MULTILINE is used to tell to Logspoud, which is used by GrayLog, to handle multiline support of the logs of the concerned container.
+# See this issue to have more informations : https://github.com/ITISFoundation/osparc-ops/issues/40
 # osparc-simcore stack (framework stack)
```
pcrespov: MINOR: Add this as a note under the title ...e.g.
```
# osparc-simcore stack (framework stack
#
# NOTES:
# - LOGSPOUT_MULTILINE is used to tell to Logspoud ... . Please add with every new osparc-simcore service.
```

services/simcore/docker-compose.deploy.yml
```diff
@@ -1,3 +1,6 @@
+
+# LOGSPOUT_MULTILINE is used to tell to Logspoud, which is used by GrayLog, to handle multiline support of the logs of the concerned container.
```
sanderegg:
```suggestion
# LOGSPOUT_MULTILINE is used to tell to Logspout, which is used by GrayLog, to handle multiline support of the logs of the concerned container.
```

services/graylog/docker-compose.yml
Surfict: Yes, i had some problems to escape the : in the end of the string. Added [:] to find a workaround,

PR #48, opened 2019-11-25
Add prefix to services

sanderegg:
## What do these changes do?

- [x] fix the infinite deployment loop (#1157)
- [x] adds option to rename the services in the simcore stack using a prefix, which effectively solves the issue of having several simcore stack deployed in the same swarm

## Checklist

- [x] **Runs in the swarm**
- [x] If you design a new module, add your user to .github/CODEOWNERS


services/deployment-agent/src/simcore_service_deployment_agent/auto_deploy_task.py
```diff
@@ -152,6 +164,10 @@
     # add parameter to the stack file if needed
     stack_cfg = await add_parameters(app_config, stack_cfg)
     log.debug("new stack config is\n%s", stack_file)
+    # change services names to avoid conflicts in common networks
+    stack_cfg = await add_prefix_to_services(app_config, stack_cfg)
+    log.debug("final stack config is: %s", stack_cfg)
```
pcrespov: MINOR: to display dicts, try ``pprint.pformat``

PR #51, opened 2019-11-29
adds recipe for official certificate or no certificate

sanderegg:
## What do these changes do?

- adds a functional recipe to deploy portus without certificate

## Checklist

- [ ] **Runs in the swarm**

### DIFF
--- a/services/portus/Makefile
+++ b/services/portus/Makefile
@@ -60,6 +60,12 @@ else
 	$(error please copy your certificate to secrets/domain.crt and private key to secrets/domain.key)
 endif
 
+.PHONY: up-no-certs
+up-no-certs: .init .create-key-cert-pair .env ${TEMP_COMPOSE}  ## Deploys portainer stack using no certificates OR a valid certificate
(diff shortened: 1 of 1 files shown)


services/portus/Makefile
```diff
@@ -60,6 +60,12 @@ else
 	$(error please copy your certificate to secrets/domain.crt and private key to secrets/domain.key)
 endif
 
+.PHONY: up-no-certs
+up-no-certs: .init .create-key-cert-pair .env ${TEMP_COMPOSE}  ## Deploys portainer stack using no certificates OR a valid certificate
```
pcrespov: if possible, lets try using nouns for target names except for the classic ``up, clean, etc``. For instance, I would prefer ``make key-cert-pair`` to ``make create-key-cert-pair`` ...

sanderegg: this is a hidden recipe that make sense inside. you will not call make create-key-cert-pair.
should create a macro instead I think.

PR #55, opened 2020-01-07
Log signature change

sanderegg:
## What do these changes do?

- adds logs when docker image signature change to further debug why there are new deployment happening
- fixed issue where swarm endpoint is not 1 by adding a function to find the id of the first endpoint.

Bonus:
added github actions CI for deployment-agent (not used for checking PRs yet, with python 3.6, 3.7 and failing 3.8)

## Checklist

- [ ] **Runs in the swarm**


ci/github/helpers/build_docker_image_tag
```diff
@@ -0,0 +1,34 @@
...
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -euo pipefail
+IFS=$'\n\t'
+
+if [ ! -v GITHUB_ACTIONS ] || [ $GITHUB_ACTIONS = "false" ]; then
+    # no travis here so let's use the git name directly
+    image_tag=$(git rev-parse --abbrev-ref HEAD)
+else
+    branch_name=${GITHUB_REF##*/}
+    if [ "$branch_name" = "master" ]; then
```
Surfict: As the same code is executed in both conditions, maybe using a -o would be more convenient ?

sanderegg: that's right. since the github action does not yet deploy I left it open.

PR #63, opened 2020-04-22
Add shellcheck + docs for cleaning a docker registry

sanderegg:
## What do these changes do?

- added documentation on how to clean a registry
- added shellcheck script
-small change for traefik_api fake service, use sleep rather than top to reduce log polution

## Checklist

- [ ] **Runs in the swarm**


docs/purge-docker-registry.md
```diff
@@ -0,0 +1,48 @@
+# How-to purge docker registry
+
+This document explains the steps necessary to purge a docker registry.
+
+## Dependencies
+
```
pcrespov, line 6: Why dont' you just define the steps below as makefile recipes and put everything inside a  ``Dockerfile``. It would be roughly something like

```Dockerfile
FROM alpine:latest

ENV DOCKER_USERNAME
ENV DOCKER_PASSWORD
ENV REGISTRY_URL

COPY docker-registry-curl.bash /scripts
COPY Makefile /scripts/Makefile

RUN apk add --no-cache \
   curl \
   jq \
   bash

ENTRYPOINT make --directory Makefile
```
Then you wrap the whole thing as with ``shellcheck.bash``, say ``registry-pruner-cli.bash``

```bash

IMAGE_NAME=local/docker-registry-purger:latest

if not exists docker images $IMAGE_NAME
  docker build --tag $IMAGE_NAME .
fi

exec docker run --env-file=.venv --rm --interactive $IMAGE_NAME "$@"
``` 

and then

1. `` registry-pruner-cli.bash list``
1. `` registry-pruner-cli.bash list target=YourRepo``
...

pcrespov: This could even be a template to follow for all tooling!

sanderegg: Well I think a doc is a minima.
A script would be good enough, makefile +dockerfile a luxury. But it's a good point.

PR #65, opened 2020-05-19
Aws integration

Surfict:
Integrate the AWS integration in Ops. 

Major changes :  

- Portus has been replaced by a simple registry
- Service mail added
- Deployment script on AWS added
- The env variable are now centralized on the repo.config and dispatched automatically in the different env files when the script is launched

## Checklist

- [ ] **Runs in the swarm**


scripts/aws-deploy.sh
```diff
@@ -0,0 +1,288 @@
...
+current_git_branch=$(git rev-parse --abbrev-ref HEAD)
+
+
+# Loads configurations variables
+# See https://askubuntu.com/questions/743493/best-way-to-read-a-config-file-in-bash
+source ${repo_basedir}/repo.config
+
+# Mails host and login/password
+
+echo $SMTP_HOST | grep -Eo "([^.*.*]+)"
```
sanderegg, line 28: ensure you shellcheck that file for lint errors

repo.config
```diff
@@ -1,7 +1,79 @@
...
-MACHINE_FQDN=osparc.local
+
 SERVICES_USER=admin
 SERVICES_PASSWORD=adminadmin
+# the machine fully qualified domain name
+MACHINE_FQDN=osparc.local
+MONITORING_DOMAIN=monitoring.${MACHINE_FQDN}
+PORTAINER_DOMAIN=portainer.${MACHINE_FQDN}
+STORAGE_DOMAIN=storage.${MACHINE_FQDN}
+REGISTRY_DOMAIN=registry.${MACHINE_FQDN}
```
sanderegg, line 12: how do you substitute these entries? using envsubst?

scripts/aws-deploy.sh
```diff
@@ -0,0 +1,288 @@
...
+$psed -i -e "s/RABBIT_PORT=.*/RABBIT_PORT=$RABBIT_PORT/" .env
+$psed -i -e "s/RABBIT_HOST=.*/RABBIT_HOST=$RABBIT_HOST/" .env
+$psed -i -e "s/RABBIT_LOG_CHANNEL=.*/RABBIT_LOG_CHANNEL=$RABBIT_LOG_CHANNEL/" .env
+$psed -i -e "s/RABBIT_PROGRESS_CHANNEL=.*/RABBIT_PROGRESS_CHANNEL=$RABBIT_PROGRESS_CHANNEL/" .env
+$psed -i -e "s/RABBIT_USER=.*/RABBIT_USER=$RABBIT_USER/" .env
+$psed -i -e "s/RABBIT_PASSWORD=.*/RRABBIT_PASSWORD=$RABBIT_PASSWORD/" .env
+
+# Reddis
+$psed -i -e "s/REDIS_HOST=.*/REDIS_HOST=$REDIS_HOST/" .env
+$psed -i -e "s/REDIS_PORT=.*/REDIS_PORT=$REDIS_PORT/" .env
```
sanderegg, line 86: would it not be simpler to either:
1. export all the entries in repo.config? this would override the ones in .env no?

2. echo repo.config >> .env that would add all the values at the end of .env and overwrite all of them

I am pretty sure at least 2. would work. then you do not need to change that script everytime we add a variable

scripts/aws-deploy.sh
```diff
@@ -0,0 +1,288 @@
...
+
+# Traefik does not use https
+$psed -i -e 's/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.entrypoints=.*/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.entrypoints=http/' docker-compose.deploy.yml
+$psed -i -e 's/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.tls=.*/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.tls=false/' docker-compose.deploy.yml
+
+# We don't use a auto-generated root certificate for storage
+$psed -i -e 's/\s\s\s\ssecrets:/    #secrets:/' docker-compose.deploy.yml
+$psed -i -e 's/\s\s\s\s\s\s- source: rootca.crt/      #- source: rootca.crt/' docker-compose.deploy.yml
+$psed -i -e "s~\s\s\s\s\s\s\s\starget: /usr/local/share/ca-certificates/osparc.crt~        #target: /usr/local/share/ca-certificates/osparc.crt~" docker-compose.deploy.yml
+$psed -i -e 's~\s\s\s\s\s\s- SSL_CERT_FILE=/usr/local/share/ca-certificates/osparc.crt~      #- SSL_CERT_FILE=/usr/local/share/ca-certificates/osparc.crt~' docker-compose.deploy.yml
```
sanderegg, line 100: prefer composition.
you can easily create a docker-compose.aws.yml file where these changes are written instead of using sed which will fail as soon as we a small change in the syntax.

scripts/aws-deploy.sh
```diff
@@ -0,0 +1,288 @@
...
+
+# -------------------------------- PORTAINER ------------------------------
+echo
+echo -e "\e[1;33mstarting portainer...\e[0m"
+pushd ${repo_basedir}/services/portainer
+$psed -i -e "s/MACHINE_FQDN=.*/MACHINE_FQDN=$MACHINE_FQDN/" .env
+$psed -i -e "s/MONITORING_DOMAIN=.*/MONITORING_DOMAIN=$MONITORING_DOMAIN/" .env
+make up-aws
+popd
+
```
sanderegg, line 122: for later... I think we should switch to jinja2 or use envsubst instead of these fragile sed stuff..

scripts/local-deploy.sh
```diff
@@ -4,16 +4,22 @@
...
+
 # Using osx support functions
 # shellcheck source=/dev/null
 source "$( dirname "${BASH_SOURCE[0]}" )/portable.sh"
 # ${psed:?}
 set -euo pipefail
 IFS=$'\n\t'
+# Are we using WSL ?
+grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null
+is_WSL=$?
```
sanderegg, line 17: is this really needed?

scripts/local-deploy.sh
```diff
@@ -60,12 +66,99 @@ cd "$repo_basedir";
...
+$psed -i -e "s/RABBIT_LOG_CHANNEL=.*/RABBIT_LOG_CHANNEL=$RABBIT_LOG_CHANNEL/" .env
+$psed -i -e "s/RABBIT_PROGRESS_CHANNEL=.*/RABBIT_PROGRESS_CHANNEL=$RABBIT_PROGRESS_CHANNEL/" .env
+$psed -i -e "s/RABBIT_USER=.*/RABBIT_USER=$RABBIT_USER/" .env
+$psed -i -e "s/RABBIT_PASSWORD=.*/RABBIT_PASSWORD=$RABBIT_PASSWORD/" .env
+
+# Reddis
+$psed -i -e "s/REDIS_HOST=.*/REDIS_HOST=$REDIS_HOST/" .env
+$psed -i -e "s/REDIS_PORT=.*/REDIS_PORT=$REDIS_PORT/" .env
+
+
```
sanderegg, line 128: same as for the aws version...
this is very fragile

scripts/local-deploy.sh
```diff
@@ -78,89 +171,70 @@ cp "${repo_basedir}"/certificates/*.crt secrets/
...
-)
-    curl -H "Accept: application/json" -H "Content-Type: application/json" -H "Portus-Auth: $SERVICES_USER:${portus_token}"  -X POST \
-        -d "${json_data}" https://"$MACHINE_FQDN":5000/api/v1/registries
-fi
+$psed -i -e "s/REGISTRY_DOMAIN=.*/REGISTRY_DOMAIN=$REGISTRY_DOMAIN/" .env
+$psed -i -e "s/S3_ACCESS_KEY_ID=.*/S3_ACCESS_KEY_ID=$SERVICES_PASSWORD/" .env
+$psed -i -e "s/S3_SECRET_ACCESS_KEY=.*/S3_SECRET_ACCESS_KEY=$SERVICES_PASSWORD/" .env
+$psed -i -e "s/S3_BUCKET=.*/S3_BUCKET=${S3_BUCKET}/" .env
+$psed -i -e "s/S3_ENDPOINT=.*/S3_ENDPOINT=${S3_ENDPOINT}/" .env
+make up
```
sanderegg, line 211: yes let's get rid of portus!!!

scripts/local-deploy.sh
```diff
@@ -78,89 +171,70 @@ cp "${repo_basedir}"/certificates/*.crt secrets/
...
+    then
+        $psed -i -e "s~- /etc/hostname:/etc/nodename # don't work with windows~#- /etc/hostname:/etc/nodename # don't work with windows~" docker-compose.yml
+    fi
+else
+    if [ $(grep  "#- /etc/hostname:/etc/nodename # don't work with windows" &> /dev/null docker-compose.yml) ]  
+    then
+        $psed -i -e "s~#- /etc/hostname:/etc/nodename # don't work with windows~- /etc/hostname:/etc/nodename # don't work with windows~" docker-compose.yml
+    fi
+fi
+
```
sanderegg, line 237: is that why you have the is_WSL? I would not bother about windows...

sanderegg: I am not sure why you have an else case ?? actually this is going to be not necessary with WSL2...

services/Makefile
```diff
@@ -0,0 +1,56 @@
...
+		,                 \
+		docker swarm init \
+	)
+	@$(if $(filter $(PUBLIC_NETWORK), $(shell docker network ls --format="{{.Name}}")) \
+		, docker network ls --filter="name==$(PUBLIC_NETWORK)" \
+		, docker network create --attachable --driver=overlay --subnet=10.10.0.0/16 $(PUBLIC_NETWORK) \
+	)
+	@$(if $(filter $(MONITORED_NETWORK), $(shell docker network ls --format="{{.Name}}")) \
+		, docker network ls --filter="name==$(MONITORED_NETWORK)" \
+		, docker network create --attachable --driver=overlay --subnet=10.11.0.0/16 $(MONITORED_NETWORK) \
```
sanderegg, line 55: why is the subnet fixed? are you sure this is always available? why not let the swarm choose it?

services/adminer/.env
```diff
@@ -1 +1,2 @@
 MONITORING_DOMAIN=monitoring.osparc.local
+PG_HOST=
```
sanderegg, line 2: rename POSTGRES_DEFAULT_SERVER=postgres

services/deployment-agent/deployment_config.default.yaml
```diff
@@ -20,7 +20,7 @@ main:
         - services/docker-compose.yml
         - .env-devel
     - id: simcore-ops-repo
-      url: https://github.com/ITISFoundation/osparc-ops.git
+      url: https://github.com/Surfict/osparc-ops.git
```
sanderegg, line 23: revert

services/deployment-agent/deployment_config.default.yaml
```diff
@@ -50,11 +50,14 @@ main:
...
-    excluded_services: [webclient, minio]
+    excluded_services: [webclient]
     excluded_volumes: []
     additional_parameters:
       extra_hosts: {}
-      environment: {}
+      environment:
+        S3_ENDPOINT: storage.osparc.local
+        S3_ACCESS_KEY: adminadmin
+        S3_SECRET_KEY: adminadmin
```
sanderegg, line 60: revert

services/mail/Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+# Makefile including function used by every services
+include ../Makefile
+
+.PHONY: up-aws
+up-aws: .env ${TEMP_COMPOSE}  ## Deploys jaeger stack in aws
+	docker stack deploy -c ${TEMP_COMPOSE} ${STACK_NAME}
+
+
+# Helpers -------------------------------------------------
```
sanderegg, line 25: that is great.
by the way we might also need it for non-aws deploys

services/mail/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    restart: always
+    networks:
+      - public
+      - monitored
+    deploy:
+      labels:
+        - traefik.enable=true
+        - traefik.docker.network=public-network
+        - traefik.http.services.mail.loadbalancer.server.port=25
+        - traefik.http.routers.mail.rule=Host(`smtp.simcore.io`)
```
sanderegg, line 33: shouldn't it be an env variable?

services/mail/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    - SYS_PTRACE
+    restart: always
+    networks:
+      - public
+      - monitored
+    deploy:
+      labels:
+        - traefik.enable=true
+        - traefik.docker.network=public-network
+        - traefik.http.services.mail.loadbalancer.server.port=25
```
sanderegg, line 32: env variable?

services/mail/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    - NET_ADMIN
+    - SYS_PTRACE
+    restart: always
+    networks:
+      - public
+      - monitored
+    deploy:
+      labels:
+        - traefik.enable=true
+        - traefik.docker.network=public-network
```
sanderegg, line 31: should be:
```
- traefik.docker.network=${PUBLIC_NETWORK}
```

services/mail/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+    driver: local
+  maillogs:
+    driver: local
+networks:
+  public:
+    external: true
+    name: public-network
+  monitored:
+    name: monitored_network
+    external: true
```
sanderegg, line 50: should be:
```
 monitored:
    name: ${MONITORED_NETWORK}
    external: true
  public:
    external: true
    name: ${PUBLIC_NETWORK}
```

sanderegg: please check for example in the adminer docker-compose/makefile files

services/mail/env-mailserver
```diff
@@ -0,0 +1,347 @@
+
```
sanderegg, line 1: why not just .env?

services/mail/setup.sh
```diff
@@ -0,0 +1,324 @@
+#!/bin/sh
+
+##
+# Wrapper for various setup scripts included in the docker-mailserver
+#
+
+CRI=
+
```
sanderegg, line 8: no strict rule enforcing?
for sh it's:
```
#!/bin/sh
set -o errexit
set -o nounset

IFS=$(printf '\n\t')
```

services/monitoring/Makefile
```diff
@@ -17,63 +17,32 @@ TEMP_COMPOSE=.stack.${STACK_NAME}.yaml
...
-MONITORED_NETWORK = monitored_network
-else
-MONITORED_NETWORK := $(monitored_network)
-endif
 
 # postgres db ulr
 POSTGRES_EXPORTER_DATA_SOURCE_NAME ?= postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${POSTGRES_DB}?sslmode=disable
 
 export MONITORED_NETWORK
 export POSTGRES_EXPORTER_DATA_SOURCE_NAME
```
sanderegg, line 25: is that not in the common MAkefile already?

services/monitoring/docker-compose.aws.yml
```diff
@@ -0,0 +1,24 @@
...
+services:
+  prometheus:
+    command:
+      - '--config.file=/etc/prometheus/prometheus.yml'
+      - '--storage.tsdb.path=/prometheus'
+      - '--storage.tsdb.retention=${PROMETHEUS_RETENTION:-24h}'
+      - '--web.console.libraries=/usr/share/prometheus/console_libraries'
+      - '--web.console.templates=/usr/share/prometheus/consoles'
+      - '--web.external-url=https://${MONITORING_DOMAIN}/prometheus/'
+      - '--web.route-prefix=/'
```
sanderegg, line 11: why do you need to override the command?

services/monitoring/docker-compose.yml
```diff
@@ -100,7 +100,7 @@ services:
       - /proc:/host/proc:ro
       - /sys:/host/sys:ro
       - /:/rootfs:ro
-      - /etc/hostname:/etc/nodename
+      - /etc/hostname:/etc/nodename # don't work with windows
```
sanderegg, line 103:
```suggestion
      - /etc/hostname:/etc/nodename # doesn't work with windows
```

services/monitoring/grafana/config.monitoring
```diff
@@ -1,4 +1,5 @@
 GF_SECURITY_ADMIN_PASSWORD=adminadmin
 GF_USERS_ALLOW_SIGN_UP=false
 GF_SERVER_ROOT_URL=https://osparc.local/grafana
+GF_SERVER_DOMAIN=
```
sanderegg, line 4: what is this one for?

services/portainer/.env
```diff
@@ -1,3 +1,4 @@
 PORTAINER_ADMIN_PWD=adminadmin
 # for let's encrypt use case
+MACHINE_FQDN=osparc.local
```
sanderegg, line 3: needed here?

services/portainer/docker-compose.yml
```diff
@@ -26,9 +26,10 @@ services:
     command: >
       -H tcp://tasks.agent:9001 --tlsskipverify
       --admin-password-file /run/secrets/portainer_admin_password
-    ports:
-      - "8000:8000"
-      - "9000:9000"
+      
+    # ports:
+    #   - "8000:8000"
+    #   - "9000:9000"
```
sanderegg, line 32: just remove that stuff

services/registry/.env
```diff
@@ -0,0 +1,8 @@
+REGISTRY_DOMAIN=
+
+S3_ACCESS_KEY_ID=
+S3_SECRET_ACCESS_KEY=
+S3_BUCKET=
+S3_ENDPOINT=
+S3_SECURE=true
```
sanderegg, line 7: put sensible default entries. don't let them empty

services/registry/docker-compose.aws.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.7'
+services:
+  registry:
+      image: registry:2
+      environment:
+        REGISTRY_STORAGE_REDIRECT_DISABLE: "true"
+        REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: inmemory
+        REGISTRY_STORAGE_REDIRECT_DISABLE: "true"
+        REGISTRY_STORAGE_DELETE_ENABLED: "true"
```
sanderegg, line 9: these are already set in the base docker-compose. do not duplicate.

services/registry/docker-compose.aws.yml
```diff
@@ -0,0 +1,20 @@
...
+      networks:
+          - public
```
sanderegg, line 11: also already in the base docker-compose

services/registry/docker-compose.aws.yml
```diff
@@ -0,0 +1,20 @@
...
+      deploy:
+          labels:
+            - traefik.http.routers.registry.tls=false
+            - traefik.http.routers.registry.entrypoints=http
+
+networks:
+  public:
+    external: true
+    name: ${PUBLIC_NETWORK}
```
sanderegg, line 20: already in the base docker-compose.

services/registry/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+            - traefik.http.routers.registry.rule=Host(`${REGISTRY_DOMAIN}`)
+            - traefik.http.routers.registry.entrypoints=https
+            - traefik.http.routers.registry.tls=true
+            - traefik.http.routers.registry.priority=10
+            - traefik.http.routers.registry.middlewares=gzip@docker, auth@docker
+
+networks:
+  public:
+    external: true
+    name: ${PUBLIC_NETWORK}
```
sanderegg, line 38: it needs to be in the monitored network as well
please add
```
monitored:
    name: ${MONITORED_NETWORK}
    external: true
```

services/registry/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+        REGISTRY_STORAGE_S3_REGION: "us-east-1"
+        REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: inmemory
+        REGISTRY_STORAGE_REDIRECT_DISABLE: "true"
+        REGISTRY_STORAGE_DELETE_ENABLED: "true"  
+        # Metrics
+        REGISTRY_HTTP_DEBUG_ADDR: ":5001"
+        REGISTRY_HTTP_DEBUG_PROMETHEUS_ENABLED: "true"
+        REGISTRY_HTTP_DEBUG_PROMETHEUS_PATH: "/metrics"
+      networks:
+          - public
```
sanderegg, line 22: please also add
```- monitored```

File: services/registry/docker-compose.yml [hunk 33]
```diff
@@ -0,0 +1,38 @@
...
+            - traefik.enable=true
+            - traefik.docker.network=${PUBLIC_NETWORK}
+            # direct access through port
+            - traefik.http.services.registry.loadbalancer.server.port=5000
+            - traefik.http.routers.registry.rule=Host(`${REGISTRY_DOMAIN}`)
+            - traefik.http.routers.registry.entrypoints=https
+            - traefik.http.routers.registry.tls=true
+            - traefik.http.routers.registry.priority=10
+            - traefik.http.routers.registry.middlewares=gzip@docker, auth@docker
+
```
sanderegg, line 34: in the original compose some stuff are missing here:
```
configs:
      - source: registry_init
        target: /etc/docker/registry/init
        mode: 0755
      - source: nsswitch_conf
        target: /etc/nsswitch.conf
        mode: 0644
```

services/registry/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+networks:
```
sanderegg, line 35: also add the general configs
```
configs:
  registry_config.yml:
    file: ./registry/config.yml
  nsswitch_conf:
    file: ./registry/nsswitch.conf
  registry_init:
    file: ./registry/init
```

services/registry/docker-compose.self-signed.yml
```diff
@@ -36,3 +28,7 @@ secrets:
   # root authority
   rootca.crt:
     external: true
+
+configs:
+  registry_init:
+    file: ./init
```
sanderegg, line 34: should be in the base docker-compose

services/registry/docker-compose.self-signed.yml
```diff
@@ -26,6 +13,11 @@ services:
       - source: rootca.crt
         target: /secrets/rootca.crt
 
+    configs:
+      - source: registry_init
+        target: /etc/docker/registry/init
+        mode: 0755
+
```
sanderegg, line 20: move to base docker-compose

services/registry/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: '3.7'
+services:
+  registry:
+      image: registry:2
+      environment:
+        REGISTRY_HTTP_HOST: "https://${REGISTRY_DOMAIN}"
```
sanderegg, line 6: add ```UPDATE_CERTIFICATES: "false"```

services/simcore/.env
```diff
@@ -10,49 +10,53 @@ CATALOG_HOST=${PREFIX_STACK_NAME}_catalog
...
 
 OPS_MONITORED_NETWORK=monitored_network
 OPS_PUBLIC_NETWORK=public-network
 
 POSTGRES_DB=simcoredb
-POSTGRES_ENDPOINT=${PREFIX_STACK_NAME}_postgres:5432
-POSTGRES_HOST=${PREFIX_STACK_NAME}_postgres
-
+POSTGRES_ENDPOINT=simcore_postgres:5432
+POSTGRES_HOST=simcore_postgres
```
sanderegg, line 22: revert

services/simcore/.env
```diff
@@ -10,49 +10,53 @@ CATALOG_HOST=${PREFIX_STACK_NAME}_catalog
...
 POSTGRES_PASSWORD=adminadmin
 POSTGRES_PORT=5432
 POSTGRES_USER=scu
 
-PUBLISHED_HOST_NAME=staging.osparc.local
+PUBLISHED_HOST_NAME=osparc.local
```
sanderegg, line 27: once all the things are ready. we need to use jinja2 or envsusbt instead of sed.

This way we could have things like:
PUBLISHED_HOST_NAME={{ repo.config.host_name }}

services/simcore/.env
```diff
@@ -10,49 +10,53 @@ CATALOG_HOST=${PREFIX_STACK_NAME}_catalog
...
 MONITORING_DOMAIN=monitoring.osparc.local
 
 RABBIT_PORT=5672
-RABBIT_HOST=${PREFIX_STACK_NAME}_rabbit
+RABBIT_HOST=simcore_rabbit
```
sanderegg, line 31: revert

services/simcore/.env
```diff
@@ -10,49 +10,53 @@ CATALOG_HOST=${PREFIX_STACK_NAME}_catalog
...
 RABBIT_LOG_CHANNEL=comp.backend.channels.log
 RABBIT_PROGRESS_CHANNEL=comp.backend.channels.progress
 RABBIT_USER=simcore
 RABBIT_PASSWORD=simcore
 
-REDIS_HOST=${PREFIX_STACK_NAME}_redis
+REDIS_HOST=simcore_redis
```
sanderegg, line 37: revert

services/simcore/.env
```diff
@@ -10,49 +10,53 @@ CATALOG_HOST=${PREFIX_STACK_NAME}_catalog
...
 REGISTRY_PW=adminadmin
 REGISTRY_SSL=True
-REGISTRY_URL=masu.speag.com
+REGISTRY_URL=registry.osparc.local
 REGISTRY_USER=admin
 
 S3_ACCESS_KEY=adminadmin
-S3_BUCKET_NAME=${PREFIX_STACK_NAME}-simcore
-S3_ENDPOINT=osparc.local:10000
+S3_BUCKET_NAME=simcore
```
sanderegg, line 47: revert

services/traefik/docker-compose.local.yml
```diff
@@ -56,9 +50,6 @@ services:
         - traefik.http.routers.traefik_http_to_https.entrypoints=traefik
         - traefik.http.routers.traefik_http_to_https.middlewares=traefik_force_https
 
-        # traefik api
-        - traefik.http.routers.api.rule=PathPrefix(`/dashboard`) || PathPrefix(`/api`)
-        - traefik.http.routers.api.entrypoints=traefik
```
sanderegg, line 61: I think there is a misunderstanding here.
this is the external traefik dashboard. this one should be kept. please revert

repo.config
```diff
@@ -1,7 +1,79 @@
...
+GRAYLOG_ROOT_PASSWORD=adminadmin
+
+
+# Postgres
+POSTGRES_DB=simcoredb
+POSTGRES_ENDPOINT=${SIMCORE_STACK_NAME}_postgres:5432
+POSTGRES_HOST=${SIMCORE_STACK_NAME}_postgres
+
+POSTGRES_USER=scu
+POSTGRES_PASSWORD=adminadmin
```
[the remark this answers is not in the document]
Surfict: Maybe we should use admin here for the pg user ?

scripts/aws-deploy.sh
Surfict: Add that in the deployer config

services/Makefile
Surfict: common.makefile add

repo.config
```diff
@@ -1,7 +1,79 @@
 REPO_USER=foo
 REPO_PASSWORD=foo
 ENVIRONMENT_MANAGER_NODE=some_host_or_IP
```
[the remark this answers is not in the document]
Surfict: It is in the original REPO.CONFIG file (https://github.com/ITISFoundation/osparc-ops/blob/master/repo.config ). No idea that it is used for

repo.config
```diff
@@ -1,7 +1,79 @@
...
+POSTGRES_PORT=5432
+
+# registry
+REGISTRY_USER=admin
+REGISTRY_PW=adminadmin
+REGISTRY_AUTH=True
+REGISTRY_SSL=True
+REGISTRY_S3_ACCESS_KEY_ID=${ACCESS_KEY_ID}
+REGISTRY_S3_SECRET_ACCESS_KEY=${SECRET_ACCESS_KEY}
+REGISTRY_S3_BUCKET=registry.osparc.io
```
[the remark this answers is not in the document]
Surfict: In production they currently have the same bucket (registry.ospar.io)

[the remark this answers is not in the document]
sanderegg: actually you are right. for the registry that is true, but could be different in some cases.
What I meant more the the S3 bucket used by the deployed simcore. This one must be separated by deploy.

services/mail/Makefile
Surfict: I add it it my todo-list for later

Makefile
```diff
@@ -1,162 +1,186 @@
...
+.PHONY: .install-fqdn
+.install-fqdn:
+	@$(if $(IS_WSL), \
+		if ! grep -Fq "$(MACHINE_IP) $(MACHINE_FQDN)" /c/Windows/System32/drivers/etc/hosts; then \
+		echo -n "Do you wish to install the following host on the Windows host ? [y/N]" && read ans && [ $${ans:-N} = y ] &&  \
+		( echo "please run the following in a PWshell with Admin rights:" && \
+		echo "Add-Content c:\Windows\System32\drivers\etc\hosts \"\`r\`$(MACHINE_IP) $(MACHINE_FQDN)\`r\`$(MACHINE_IP) $(MONITORING_DOMAIN)\`r\`$(MACHINE_IP) $(REGISTRY_DOMAIN)\`r\`$(MACHINE_IP) $(PORTAINER_DOMAIN)\`r\`$(MACHINE_IP) $(STORAGE_DOMAIN)\"" && \
+		echo "OR please run the following in a CMD with Admin rights (note that wildcards are not accepted):" && \
+		echo "echo $(MACHINE_IP) $(MACHINE_FQDN) >> c:\Windows\System32\drivers\etc\hosts && echo $(MACHINE_IP) $(MONITORING_DOMAIN) >> c:\Windows\System32\drivers\etc\hosts && echo $(MACHINE_IP) $(PORTAINER_DOMAIN) >> c:\Windows\System32\drivers\etc\hosts && echo $(MACHINE_IP) $(REGISTRY_DOMAIN) >> c:\Windows\System32\drivers\etc\hosts && echo $(MACHINE_IP) $(STORAGE_DOMAIN) >> c:\Windows\System32\drivers\etc\hosts") \
+		|| [ 42 ]; \
```
sanderegg, line 76: || [ 42 ] ? what is this the universal number?

Makefile
```diff
@@ -1,162 +1,186 @@
...
+		fi \
+	,\
+	if ! grep -Fq "$(MACHINE_IP) $(MACHINE_FQDN)" /etc/hosts; then \
+		echo -n "Do you wish to install the following host? [y/N] " && read ans && [ $${ans:-N} = y ] && \
+		( sudo echo "$(MACHINE_IP) $(MACHINE_FQDN)\n$(MACHINE_IP) $(PORTAINER_DOMAIN)\n$(MACHINE_IP) $(REGISTRY_DOMAIN)\n$(MACHINE_IP) $(MONITORING_DOMAIN)\n$(MACHINE_IP) $(STORAGE_DOMAIN)\n" >> /etc/hosts && \
+		echo "# restarting docker daemon" && \
+		sudo systemctl restart docker ) \
+		|| [ 42]; \
```
sanderegg, line 84: same

Makefile
```diff
@@ -1,162 +1,186 @@
...
+	fi \
+	)
+
+	@$(if $(IS_WSL), \
+	if ! sudo grep -Fq "$(MACHINE_IP) $(MACHINE_FQDN)" /etc/hosts; then \
+		echo -n "Do you wish to install the following host in WSL? [y/N] " && read ans && [ $${ans:-N} = y ] && \
+		( printf  "Adding\n" && \
+		printf "$(MACHINE_IP) $(MACHINE_FQDN)\n$(MACHINE_IP) $(PORTAINER_DOMAIN)\n$(MACHINE_IP) $(REGISTRY_DOMAIN)\n$(MACHINE_IP) $(MONITORING_DOMAIN)\n$(MACHINE_IP) $(STORAGE_DOMAIN)\n" | sudo tee -a /etc/hosts && \
+		printf  "to /etc/hosts\n" ) \
+		|| [ 42 ]; \
```
sanderegg, line 94: same

README.md
```diff
@@ -31,22 +31,38 @@ cd osparc-ops
 make help
 ```
 
+### WSL deployement
+
+If you are using WSL1, clone the repository under the Windows filesystem and access it in WSL. You need to ensure [Ensure Volume Mounts Work](https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly) (Section Volume Mounts Work)
```
sanderegg, line 36: actually, maybe you should change to something like:
clone the repository in a path accessible by the Windows filesystem (typically in /mnt/c/my_path and not in /home)

If you clone using windows git, then you might get into the problem that the git in windows and wsl are not in sync...

README.md
```diff
@@ -31,22 +31,38 @@ cd osparc-ops
...
+
+Do not clone the repo directly in your WSL1 filesystem.
```
sanderegg, line 38: see above comment

services/deployment-agent/Makefile
```diff
@@ -64,6 +64,9 @@ build build-kit build-x build-devel build-devel-kit build-devel-x: ## Builds $(A
 up: .init ${DEPLOYMENT_AGENT_CONFIG} ${TEMP_COMPOSE} ## Deploys or updates current stack "$(STACK_NAME)" using replicas=X (defaults to 1)
 	@docker stack deploy -c ${TEMP_COMPOSE} $(STACK_NAME)
 
+up-aws: .init ${DEPLOYMENT_AGENT_CONFIG} ${TEMP_COMPOSE}-aws ## Deploys or updates current stack "$(STACK_NAME)" using replicas=X (defaults to 1)
```
pcrespov, line 67: Comparing ``up-aws`` with ``up-devel``, I wonder what is the different (or the correct one) between
```
${TEMP_COMPOSE}-aws
${TEMP_COMPOSE-aws}
```

README.md
```diff
@@ -31,22 +31,38 @@ cd osparc-ops
...
 A self-signed certificate may be generated. The system host file may be modified using the default **osparc.local** fully qualified domain name (FQDN) to point to the local machine.
 
 The services above will be deployed and pre-configured on the following endpoints:
-  - Traefik: [https://osparc.local:9001/dashboard/](https://osparc.local:9001/dashboard/)
-  - Portainer: [https://osparc.local/portainer/](https://osparc.local/portainer/)
-  - Minio: [https://osparc.local/minio](https://osparc.local/minio) and [https://osparc.local:10000](https://osparc.local:10000)
-  - Portus: [https://osparc.local:5000](https://osparc.local:5000)
+  - Traefik: [https://monitoring.osparc.local/dashboard/#/](https://monitoring.osparc.local/dashboard/#/)
+  - Portainer: [https://monitoring.osparc.local/portainer/](https://monitoring.osparc.local/portainer/)
+  - Minio: [https://storage.osparc.local/minio/](https://storage.osparc.local/minio/)
```
sanderegg, line 60: why is there still /minio here? in our deploys it would be just https://storage.osparc.local

README.md
```diff
@@ -31,22 +31,38 @@ cd osparc-ops
...
   make up-local
   ```
 A self-signed certificate may be generated. The system host file may be modified using the default **osparc.local** fully qualified domain name (FQDN) to point to the local machine.
 
 The services above will be deployed and pre-configured on the following endpoints:
-  - Traefik: [https://osparc.local:9001/dashboard/](https://osparc.local:9001/dashboard/)
-  - Portainer: [https://osparc.local/portainer/](https://osparc.local/portainer/)
-  - Minio: [https://osparc.local/minio](https://osparc.local/minio) and [https://osparc.local:10000](https://osparc.local:10000)
-  - Portus: [https://osparc.local:5000](https://osparc.local:5000)
+  - Traefik: [https://monitoring.osparc.local/dashboard/#/](https://monitoring.osparc.local/dashboard/#/)
```
sanderegg, line 58: are you sure about the trailing #/ ?

repo.config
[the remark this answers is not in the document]
sanderegg: well I think this is for deploying local docker-machines. it is not in use for usual deployment.

services/mail/Makefile
```diff
@@ -0,0 +1,29 @@
...
+STACK_NAME = $(notdir $(shell pwd))
+SWARM_HOSTS = $(shell docker node ls --format={{.Hostname}} 2>/dev/null)
+TEMP_COMPOSE=.stack.${STACK_NAME}.yaml
+
+
+# External VARIABLES
+include .env
+
+# TARGETS --------------------------------------------------
+include $(realpath $(CURDIR)/../../scripts/common.mk)
```
pcrespov, line 15: MINOR: should addopt a naming convention regarding common makefiles.  See ``scripts/common.mk`` and ``services/common.Makefile``. The latter, I assume is only for services, but what about the former??

services/mail/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+version: '3.7'
+services:
+  mail:
+    dns:
+      8.8.8.8
+    image: tvial/docker-mailserver:latest
```
pcrespov, line 6: I would FREEZE all versions from externals... even with SHA!

services/monitoring/Makefile
```diff
@@ -17,63 +17,31 @@ TEMP_COMPOSE=.stack.${STACK_NAME}.yaml
...
-else
-PUBLIC_NETWORK := $(public_network)
-endif
-export PUBLIC_NETWORK
 
 # TARGETS --------------------------------------------------
 include $(realpath $(CURDIR)/../../scripts/common.mk)
 
+# Makefile including function used by every services
+include ../common.Makefile
```
pcrespov, line 31: common is definitively making every Makefile more readable and making easier to spot differences

repo.config
```diff
@@ -1,7 +1,80 @@
...
+POSTGRES_PORT=5432
+
+# registry
+REGISTRY_USER=${SERVICES_USER}
+REGISTRY_PW=${SERVICES_PASSWORD}
+REGISTRY_AUTH=True
+REGISTRY_SSL=True
+REGISTRY_S3_ACCESS_KEY_ID=${ACCESS_KEY_ID}
+REGISTRY_S3_SECRET_ACCESS_KEY=${SECRET_ACCESS_KEY}
+REGISTRY_S3_BUCKET=registry.osparc.io
```
sanderegg, line 51: this should be:
REGISTRY_S3_BUCKET=${REGISTRY_DOMAIN}

services/registry/docker-compose.aws.yml
```diff
@@ -0,0 +1,10 @@
+version: '3.7'
+services:
+  registry:
+      image: registry:2
```
pcrespov, line 4: FREEZE images: in this case ``registry:2.7.1`` https://hub.docker.com/_/registry

repo.config
```diff
@@ -1,7 +1,80 @@
...
+
+# Rabbit
+RABBIT_PORT=5672
+RABBIT_HOST=${PREFIX_STACK_NAME}_rabbit
+RABBIT_LOG_CHANNEL=comp.backend.channels.log
+RABBIT_PROGRESS_CHANNEL=comp.backend.channels.progress
+RABBIT_USER=${SERVICES_USER}
+RABBIT_PASSWORD=${SERVICES_PASSWORD}
+
+#Reddis
```
sanderegg, line 64:
```suggestion
#Redis
```

scripts/fill-env.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -euo pipefail
+IFS=$'\n\t'
+
```
sanderegg, line 5: missing something here?

services/registry/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
+version: '3.7'
+services:
+  registry:
+      image: registry:2
```
pcrespov, line 4: FREEZE image version

scripts/local-deploy.sh
sanderegg, line 17:
```suggestion
# Are we using WSL ?
IS_WSL= $(grep -qEi "(Microsoft|WSL)" /proc/version &> /dev/null)
```

services/registry/docker-compose.aws.yml
pcrespov, line 4: there is no need to repeat image here since ``docker-compose.aws.yml`` is a partial config and the image is defined in the main ``docker-compose.yml``.

services/simcore/docker-compose.deploy.yml
```diff
@@ -57,12 +57,16 @@ services:
         window: 120s
 
   director:
+    secrets:
+      - source: rootca.crt
+        target: /usr/local/share/ca-certificates/osparc.crt
```
pcrespov, line 62: is this ok to be public?

scripts/local-deploy.sh
```diff
@@ -78,89 +170,70 @@ cp "${repo_basedir}"/certificates/*.crt secrets/
...
 echo -e "\e[1;33mstarting monitoring...\e[0m"
 # set MACHINE_FQDN
 pushd "${repo_basedir}"/services/monitoring
+$psed -i -e "s/MONITORING_DOMAIN=.*/MONITORING_DOMAIN=$MONITORING_DOMAIN/" .env
 $psed -i -e "s|GF_SERVER_ROOT_URL=.*|GF_SERVER_ROOT_URL=https://$MACHINE_FQDN/grafana|" grafana/config.monitoring
 $psed -i -e "s|GF_SECURITY_ADMIN_PASSWORD=.*|GF_SECURITY_ADMIN_PASSWORD=$SERVICES_PASSWORD|" grafana/config.monitoring
 $psed -i -e "s|basicAuthPassword:.*|basicAuthPassword: $SERVICES_PASSWORD|" grafana/provisioning/datasources/datasource.yml
-$psed -i -e "s|--web.external-url=.*|--web.external-url=https://$MACHINE_FQDN/prometheus/'|" docker-compose.yml
+
+# if  WSL, we comment - /etc/hostname:/etc/host_hostname
```
sanderegg, line 223: maybe can you add the comment that this must be commented if deploying on windows

services/traefik/Makefile
```diff
@@ -17,70 +17,43 @@ SHELL := $(if $(IS_WIN),powershell.exe,/bin/bash)
...
+# Makefile including function used by every services
+include ../common.Makefile
+
 .PHONY: up
 up-local: .init .create_secrets .env ${TEMP_COMPOSE}-local ## Deploys the stack using provided certificates
 	docker stack deploy -c ${TEMP_COMPOSE}-local ${STACK_NAME}
 
+.PHONY: up-aws
+up-aws: .env ${TEMP_COMPOSE}-aws ## Deploys the stack aws 
+	@echo "test"
```
pcrespov, line 42: MINOR: remove these echos
Use instead inforative comments, eg.
`` # deploying ${STACK_NAME} to aws ``

scripts/local-deploy.sh
```diff
@@ -178,13 +262,29 @@ echo -e "\e[1;33mstarting graylog...\e[0m"
 # set MACHINE_FQDN
 pushd "${repo_basedir}"/services/graylog;
 graylog_password=$(echo -n "$SERVICES_PASSWORD" | sha256sum | cut -d ' ' -f1)
-$psed -i -e "s|GRAYLOG_HTTP_EXTERNAL_URI=.*|GRAYLOG_HTTP_EXTERNAL_URI=https://$MACHINE_FQDN/graylog/|" .env
+$psed -i -e "s/MONITORING_DOMAIN=.*/MONITORING_DOMAIN=$MONITORING_DOMAIN/" .env
+$psed -i -e "s|GRAYLOG_HTTP_EXTERNAL_URI=.*|GRAYLOG_HTTP_EXTERNAL_URI=https://$MONITORING_DOMAIN/graylog/|" .env
 $psed -i -e "s|GRAYLOG_ROOT_PASSWORD_SHA2=.*|GRAYLOG_ROOT_PASSWORD_SHA2=$graylog_password|" .env
+
+# if  WSL, we comment - /etc/hostname:/etc/host_hostname
```
sanderegg, line 269: same as above

services/adminer/.env
```diff
@@ -1 +1,2 @@
 MONITORING_DOMAIN=monitoring.osparc.local
+POSTGRES_DEFAULT_SERVER=postgres
```
sanderegg, line 2: hmm not sure you sed it in the script above. this should be equal to ${POSTGRES_HOST}

services/adminer/Makefile
```diff
@@ -39,29 +26,11 @@ up: .init .env ${TEMP_COMPOSE}  ## Deploys adminer stack
 up-letsencrypt: .init .env ${TEMP_COMPOSE}-letsencrypt  ## Deploys adminer stack using let's encrypt
 	docker stack deploy -c ${TEMP_COMPOSE}-letsencrypt ${STACK_NAME}
 
-.PHONY: down
-down: ## Removes the stack from the swarm
-	docker stack rm ${STACK_NAME}
+.PHONY: up-aws
+up-aws: .init .env ${TEMP_COMPOSE}-aws  ## Deploys adminer stack using let's encrypt
```
sanderegg, line 30: let's encrypt for aws?

services/deployment-agent/Makefile
```diff
@@ -64,6 +64,9 @@ build build-kit build-x build-devel build-devel-kit build-devel-x: ## Builds $(A
...
+	@docker stack deploy -c ${TEMP_COMPOSE}-aws $(STACK_NAME)
+
```
sanderegg, line 69: out of curiosity, do you really need to have a specific recipe for aws for the deployer? it does not need any certificate

services/deployment-agent/docker-compose-aws.yml
```diff
@@ -0,0 +1,5 @@
+version: "3.7"
+services:
+  auto-deployment-agent:
+    dns :
+      - 8.8.8.8
```
sanderegg, line 5: I don't really understand why you need this for that service? is that everywhere?

services/monitoring/docker-compose.aws.yml
```diff
@@ -0,0 +1,16 @@
+version: '3.7'
+services:
+  prometheus:
+    deploy:
+      labels:
+        - traefik.http.routers.prometheus.rule=Host(`${MONITORING_DOMAIN}`) && PathPrefix(`/prometheus`)
```
sanderegg, line 6: this is not already in the base docker-compose?

services/monitoring/docker-compose.aws.yml
```diff
@@ -0,0 +1,16 @@
...
+        - traefik.http.routers.prometheus.tls=false
+        - traefik.http.routers.prometheus.entrypoints=http
+
+  grafana:
+    deploy:
+      labels:
+        # direct access through port
+        - traefik.http.routers.grafana.rule=Host(`${MONITORING_DOMAIN}`) && PathPrefix(`/grafana`)
```
sanderegg, line 14: same here

services/monitoring/docker-compose.yml
```diff
@@ -196,6 +196,8 @@ services:
           memory: 64M
 
   grafana:
+    dns:
+      8.8.8.8
```
sanderegg, line 200: This dns entry here. why is that?
I've seen you have that in the deployer docker-compose.aws.yml and here it is in the base docker-compose. any reason?

sanderegg: it looks like it is only in grafana, adminer, deployer... why not graylog?? is this really necessary? or are these remainders of tests? if so please remove them.

services/registry/docker-compose.aws.yml
sanderegg: these should be removed.
Also fix the version of the registry (currently I think it's 2.7.x)

services/registry/docker-compose.yml
sanderegg, line 4: 2.7.0 or something like this, check in dockerhub

services/simcore/.env
```diff
@@ -6,56 +6,60 @@
...
 
 OPS_MONITORED_NETWORK=monitored_network
 OPS_PUBLIC_NETWORK=public-network
 
 POSTGRES_DB=simcoredb
-POSTGRES_ENDPOINT=${PREFIX_STACK_NAME}_postgres:5432
-POSTGRES_HOST=${PREFIX_STACK_NAME}_postgres
-
+POSTGRES_ENDPOINT=local_postgres:5432
+POSTGRES_HOST=local_postgres
```
sanderegg, line 22: can we please then keep it as? so it is still compatible with the deployer in case of // deployment? You anyway replace them with the psed afterwards. this way it works with both.... same for all the ENVs below...
${PREFIX_STACK_NAME}_postgres:...

services/simcore/.env
```diff
@@ -6,56 +6,60 @@
...
-STORAGE_ENDPOINT=${PREFIX_STACK_NAME}_storage:8080
-STORAGE_HOST=${PREFIX_STACK_NAME}_storage
+SMTP_HOST=smtp.osparc.local
+SMTP_PORT=25
+SMTP_USERNAME=admin
+SMTP_PASSWORD=adminadmin
+SMTP_TLS_ENABLED=1
+
+STORAGE_ENDPOINT=${SIMCORE_STACK_NAME}_${PREFIX_STACK_NAME}_storage:8080
+STORAGE_HOST=${SIMCORE_STACK_NAME}_${PREFIX_STACK_NAME}_storage
```
sanderegg, line 62: this looks weird... are you sure it should not be just ${PREFIX_STACK_NAME}_storage ???

Makefile
Surfict: It has to be something that need to be true and I in that case I like to put 42, because why not

GitHK: Please consider readability for outcomes. They might not understand not be familiar with bash so using some form of more conventional "true" value might be easier for the rest.

services/monitoring/docker-compose.yml
Surfict: This a problem that I met several times within AWS, some container doesn't access internet if I don't set the DNS myself. True that I should that graylog could have it also one day. I will add it

repo.config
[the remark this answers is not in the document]
Surfict: I moved them in a deploy.config file.

services/deployment-agent/Makefile
Surfict: To follow the convention, the last one is the good one. I corrected

services/mail/Makefile
Surfict: Renamed it in common.makefile

Surfict: The script one add some general things like "info, help, clean" but need a little bit of refactoring, because somes recipe like clean are present in the services one.

README.md
Surfict: I don't understand your last sentence. What would not be in sync ? Are they not reading the sames files in the folder .git ?

Surfict: Ah ok maybe there is a misunderstanding here. If it's important to clone it in the windows filesystem, this is to ensure volume will be mount (they are not accessible by the windows docker engine when the folder of the project is /home)

sanderegg: ok so just for your info...
using WSL2 it is important to do the exact contrary... so I would remove this stuff or enhance.

services/monitoring/docker-compose.yml
sanderegg: but how come it happens only sometimes? it must be something else don't you think?

Makefile
Surfict: Well that's the only fantasy that I allows to myself, and I would like to keep it like this, but I will add a comment to make it easier.

services/monitoring/docker-compose.yml
Surfict: I didn't find the cause. I was thinking first it would depend of the node it is deployed but not at all. Sometimes, just restarting it it would work and some other times no. I believe this is because the DNS from the base DNS (from ubuntu) are not correctly copied into the container. But I would need to take a deeper look to understand why it happens only sometimes..

services/deployment-agent/Makefile
Surfict: The only reason is this DNS problem. I need to add it to the deployer..

Surfict: It's in my todo list to look more where this problem is coming from but for now, I choosed the easy solution with that. Sure I will be able to remove it later when this problem will be solved

README.md
Surfict: https://storage.osparc.local automatically redirect to https://storage.osparc.local/minio

README.md
Surfict: yes

services/simcore/docker-compose.deploy.yml
Surfict: The certificates itself is not push in the repo, it's only when it is generated that it will be available, so I think yes ?

repo.config
Surfict: From what I remind. the name is always registry.osparc.io, so not always ${REGISTRY_DOMAIN} no ?

sanderegg: so I just double-checked on master. there it is osparc01.speag.com for example.
it would not really make sense to always name it with osparc.io. for example in aws the registry is with simcore.io right?

services/monitoring/docker-compose.yml
sanderegg: well if you are sure this does not have ANY side effect you can probably set it everywhere. but no one will understand if it is only in some case there. Also, if you do so, add a note next to it. Or better find out why it is needed, and if it is not, remove it.

scripts/deployements/local-deploy.sh
```diff
@@ -0,0 +1,370 @@
...
+$psed -i -e "s~\s\s\s\s\s\s\s\s#target: /usr/local/share/ca-certificates/osparc.crt~        target: /usr/local/share/ca-certificates/osparc.crt~" docker-compose.deploy.yml
+$psed -i -e 's~\s\s\s\s\s\s#- SSL_CERT_FILE=/usr/local/share/ca-certificates/osparc.crt~      - SSL_CERT_FILE=/usr/local/share/ca-certificates/osparc.crt~' docker-compose.deploy.yml
+
+new_compose_simcore=$(cat docker-compose.deploy.yml)
+new_env_simcore=$(cat .env)
+if [ "$ori_env_simcore" = "$new_env_simcore" ] && [ "$ori_compose_simcore" = "$new_compose_simcore" ]; then
+    echo "Simcore service ready for deployment"
+else
+    echo "Some values weren't up to to date in Simcore service (.env and/or docker-compose.deploy). They have been updated, please push them to github and restart the script"
+    exit
```
sanderegg, line 149: if I call ```make up-local``` twice in a row this test will fail.

README.md
sanderegg: ok

scripts/deployements/aws-deploy.sh
```diff
@@ -0,0 +1,234 @@
...
+pushd "${repo_basedir}"/services/simcore;
+
+simcore_env=".env"
+simcore_compose="docker-compose.deploy.yml"
+
+substitute_environs ${simcore_env}
+
+# for aws use we need http for the traefik entrypoint in simcore. Https is handled by aws
+$psed --in-place --expression='s/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.entrypoints=.*/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.entrypoints=http/' ${simcore_compose}
+$psed --in-place --expression='s/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.tls=.*/traefik.http.routers.${PREFIX_STACK_NAME}_webserver.tls=false/' ${simcore_compose}
```
sanderegg, line 64: for later: could also be parametrized.

services/deployment-agent/.env
```diff
@@ -7,4 +7,6 @@ SIMCORE_STACK_NAME=simcore
 # defines the dockerhub registry organisation
 SIMCORE_DOCKER_REGISTRY=itisfoundation
 # set the image tag to be used from dockerhub
-SIMCORE_IMAGE_TAG=master-github-latest
+SIMCORE_IMAGE_TAG=staging-latest
```
sanderegg, line 10: should be set to ```master-github-latest```

services/deployment-agent/Makefile
```diff
@@ -77,6 +80,13 @@ down: ## Stops and remove stack from swarm
 leave: ## leaves swarm stopping all stacks, secrets in it
 	-docker swarm leave -f
 
+.PHONY: env-subst
+env-subst:
```
sanderegg, line 84:
```suggestion
.env: template.env
```
in GNU Make, this means it will check template.env exists. and also it will create/update the .env file .

CODEOWNERS
```diff
@@ -1,6 +1,6 @@
-# Order is important. The last matching pattern has the most precedence.
-
-# files and folders recursively
-/virtual_cluster/              @ehzastrow
-/scripts/                      @pcrespov
-/services/                     @sanderegg,@pcrespov
+# Order is important. The last matching pattern has the most precedence.
```
pcrespov, line 1: This entire file belongs in ``.github`` folder not here

Makefile
```diff
@@ -41,14 +42,19 @@ certificates/domain.key:
 
 .PHONY: up-local
 up-local: .install-fqdn certificates/domain.crt certificates/domain.key .create-secrets ## deploy osparc ops stacks and simcore
-	bash scripts/local-deploy.sh
+	bash scripts/deployements/local-deploy.sh
```
pcrespov, line 45: ``deployEments`` or ``deployments`` (w/o E)?

Surfict: That was a nice frenglish word...

services/traefik/docker-compose.yml
```diff
@@ -2,6 +2,8 @@ version: "3.7"
 
 services:
   traefik:
+    dns: 
+      8.8.8.8
```
sanderegg, line 6: aws only?

services/traefik/docker-compose.yml
```diff
@@ -75,6 +77,8 @@ services:
       - public
       - monitored
   whoami:
+    dns: 
+      8.8.8.8
```
sanderegg, line 81: aws only?

Makefile
```diff
@@ -13,75 +13,90 @@ IS_OSX  := $(filter Darwin,$(shell uname -a))
...
-		pushd $$service; $(MAKE) down; popd; \
+		$(MAKE_C) $$service down; \
 	done
 
 .PHONY: leave
 leave: ## leaves the swarm
 	docker swarm leave -f
 
 .PHONY: .install-fqdn
+ # [ 42 ]; = always true
```
pcrespov, line 70: Please remove all comments if not needed? is thisone necessary?

Makefile
```diff
@@ -13,75 +13,90 @@ IS_OSX  := $(filter Darwin,$(shell uname -a))
...
 
+	@$(if $(IS_WSL), \
+	if ! sudo grep -Fq "$(MACHINE_IP) $(MACHINE_FQDN)" /etc/hosts; then \
+		echo -n "Do you wish to install the following host in WSL? \n [y/N] " && read ans && [ $${ans:-N} = y ] && \
+		( printf  "Adding\n" && \
+		printf "$(MACHINE_IP) $(MACHINE_FQDN)\n$(MACHINE_IP) $(PORTAINER_DOMAIN)\n$(MACHINE_IP) $(REGISTRY_DOMAIN)\n$(MACHINE_IP) $(MONITORING_DOMAIN)\n$(MACHINE_IP) $(STORAGE_DOMAIN)\n$(MACHINE_IP) $(API_DOMAIN)\n" | sudo tee -a /etc/hosts && \
+		printf  "to /etc/hosts\n" ) \
+		|| true; \
+	fi \
+	,)
```
pcrespov, line 99: MINOR: leave spaces between recipes. it is more readable

scripts/deployments/.README
```diff
@@ -0,0 +1 @@
+# Scripts for a full deployment of ops and simcore in local and aws. Called in the root Makefile
```
pcrespov, line 1: STYLE: Why ``.README``?? and not simply ``README.md`` as usual?

scripts/create-s3-bucket.bash
```diff
@@ -0,0 +1,16 @@
...
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -o errexit
+set -o nounset
+set -o pipefail
+IFS=$'\n\t'
+
+docker run \
+-v /etc/ssl/certs:/etc/ssl/certs:ro \
+--network host \
+-e MC_HOST_local="https://${S3_ACCESS_KEY}:${S3_SECRET_KEY}@${S3_ENDPOINT}" \
```
pcrespov, line 15: MINOR: as a general rule, we agreed that when written in a script, we should use long CLI arguments, e.g. here you should use ``--environment`` instead of ``-e``

PR #69, opened 2020-06-17
Maintenance/changes from master

sanderegg:
## What do these changes do?

- sslheader middleware is not happy when there are hyphens in its name
- for local deploys a port 10081 is opened until the AWS PR comes in

## Checklist

- [ ] **Runs in the swarm**


services/simcore/.env
```diff
@@ -58,9 +58,11 @@ STORAGE_ENDPOINT=${PREFIX_STACK_NAME}_storage:8080
 STORAGE_HOST=${PREFIX_STACK_NAME}_storage
 
 SWARM_STACK_NAME=${PREFIX_STACK_NAME}-simcore
+SWARM_STACK_NAME_NO_HYPHEN=${PREFIX_STACK_NAME}_simcore
```
pcrespov, line 61: what if ``PREFIX_STACK_NAME`` has hyphen? Cannot we use [shell parameter expansions](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) in the makefile for ``PREFIX_STACK_NAME``?

services/deployment-agent/.env
```diff
@@ -7,4 +7,4 @@ SIMCORE_STACK_NAME=simcore
 # defines the dockerhub registry organisation
 SIMCORE_DOCKER_REGISTRY=itisfoundation
 # set the image tag to be used from dockerhub
-SIMCORE_IMAGE_TAG=master-latest
+SIMCORE_IMAGE_TAG=master-github-latest
```
pcrespov, line 10: why the ``gitbhu`` reference? As a default to this repo vs the environment repos in gitlab?

sanderegg: that are the images generated by github actions

services/simcore/.env
sanderegg: yes that is a good point. but normally these are named as master, staging or production

PR #73, opened 2020-09-01
Fix multiple tags same commit

sanderegg:
## What do these changes do?

in the new release process there might be several tags on the same commit.
git does not support this well as it ```git describe --tags``` returns the first tag it finds instead of all of them.

## Checklist

- [ ] **Runs in the swarm**


GitHK (approved):
managed to find those tiny changes

mguidon (approved):

PR #81, opened 2022-12-23
🐛 Fix/grafana node dashboard for autoscaling

mrnicegyu11:
This fixes the grafana "Nodes - Detailed Insights" Dashboard and makes it work even with dynamically added ("autoscaled") machines.


sanderegg (approved):

sanderegg:
I like that way of changing OPS! 💯

PR #84, opened 2023-01-04
Change pg_backup tool version from justb4/pgbackup to kartoza/pg-backup

Surfict:
With the migration to the last version of PG, our old pg-backup wasn't able to do its job anymore. This PR uses another version of PG-backup (kartoza/pg-backup) which is compatible with  V14.X. of PostgreSQL
Unlike the older version, this version will not detect automatically pg docker instances. Changes made in this PR make it ready to work for each simcore stack deployed - but it has to be deployed each time a simcore stack is deployed, unlike the older version who was taking care of the backup of all the stack deployed.


services/pg-backup/template.env
```diff
@@ -0,0 +1,5 @@
+MACHINE_FQDN=${MACHINE_FQDN}
+POSTGRES_USER=${POSTGRES_USER}
+POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+POSTGRES_PORT=${POSTGRES_PORT}
+PREFIX_STACK_NAME=${PREFIX_STACK_NAME}
```
mrnicegyu11, line 5: you didnt run the pre-commit hooks, please do so:|
`pre-commit install && pre-commit run --all`

services/pg-backup/Makefile
```diff
@@ -26,15 +26,22 @@ up-master: up-dalco
...
+	source $(REPO_CONFIG_LOCATION); \
+	sed -i "s/SERVICE_NAME_PARAMETRIZED/$${PREFIX_STACK_NAME}_pg_backup/g" $@;\
+	set +o allexport; \
 
 .PHONY: ${TEMP_COMPOSE}-public
 ${TEMP_COMPOSE}-public: docker-compose.yml docker-compose.public.yml
 	docker-compose -f $< -f docker-compose.public.yml --log-level=ERROR config > $@
+	set -o allexport; \
+	source $(REPO_CONFIG_LOCATION); \
+	sed -i "s/SERVICE_NAME_PARAMETRIZED/$${PREFIX_STACK_NAME}_pg_backup/g" $@;\
```
mrnicegyu11, line 46: why not use jinja2 here? Much more modern and powerful and less error prone. 

If you need an example on how to generate a valid docker yml file from a j2 template, see:
https://github.com/ITISFoundation/osparc-ops-environments/blob/14569774c02307fc508fceb5e967337c9a924107/services/traefik/Makefile#L104

services/pg-backup/Makefile
```diff
@@ -12,8 +12,8 @@ REPO_BASE_DIR := $(shell git rev-parse --show-toplevel)
 include ${REPO_BASE_DIR}/scripts/common.Makefile
 
 .PHONY: up ## Deploys pg-backup stack
-up: .init ${TEMP_COMPOSE}  ## Deploys pg-backup stack
-	docker stack deploy --with-registry-auth --prune --compose-file ${TEMP_COMPOSE} ${STACK_NAME}
+up: .init .env ${TEMP_COMPOSE}  ## Deploys pg-backup stack
+	docker stack deploy --with-registry-auth --prune --compose-file ${TEMP_COMPOSE} ${STACK_NAME}; \
```
mrnicegyu11, line 16: Why the trailing backslash?

services/pg-backup/Makefile
```diff
@@ -26,15 +26,22 @@ up-master: up-dalco
 
 .PHONY: up-public ## Deploys pg-backup stack for public access cluster
 up-public: .init ${TEMP_COMPOSE}-public
-	docker stack deploy --with-registry-auth --prune --compose-file ${TEMP_COMPOSE}-public ${STACK_NAME}
-
+	docker stack deploy --with-registry-auth --prune --compose-file ${TEMP_COMPOSE}-public ${STACK_NAME}; \
```
mrnicegyu11, line 29: Why the trailing backslash?

services/pg-backup/docker-compose.yml
```diff
@@ -1,14 +1,20 @@
...
+  SERVICE_NAME_PARAMETRIZED:
+    image: kartoza/pg-backup:14-3.3@sha256:4d1867707b1da879324199fb3c0d542122113bfc50cd2418764efb8f4309d6c1
     volumes:
-    - /var/run/docker.sock:/var/run/docker.sock
-    - /tank:/backup
+      - /tank/${MACHINE_FQDN}:/backups
     environment:
-    - PGB_SCHEDULE=0 * * * * # Every hours
+      - DUMPPREFIX=PG_${MACHINE_FQDN}
+      - DUMP_ARGS="--format=c"
```
mrnicegyu11, line 9: What does this mean, what is the resulting format? Let's try to stick with vanilla .sql files if possible is my suggestion

services/pg-backup/docker-compose.yml
```diff
@@ -1,14 +1,20 @@
 version: '3.7'
 services:
-  dbbackup:
-    image: justb4/pgbackup:10
+  SERVICE_NAME_PARAMETRIZED:
+    image: kartoza/pg-backup:14-3.3@sha256:4d1867707b1da879324199fb3c0d542122113bfc50cd2418764efb8f4309d6c1
```
mrnicegyu11, line 4: thanks for locking the version :)

mrnicegyu11 (changes requested):
Already looks very promising. Please have a a look at the comments and address them :)
Until we have a proper idea how to integrate this for all platforms, please add a small README.md that clearly says that the backup is not started by the Makefile of the osparc-ops-environments repo, to avoid confusion.

Thanks!

services/pg-backup/docker-compose.yml
```diff
@@ -1,14 +1,20 @@
...
     volumes:
-    - /var/run/docker.sock:/var/run/docker.sock
-    - /tank:/backup
+      - /tank/${MACHINE_FQDN}:/backups
```
mrnicegyu11, line 6: What happens if the folder /tank/${MACHINE_FQDN} doesnt exist? Will it be created or will the service fail?

Surfict: Service fail - it need to be created each time we have a new deployement, or updated each time we change this variable

services/pg-backup/docker-compose.yml
Surfict: SQL dump give something 7 times bigger (11M with the custom format, around 77M with vanilla). You will have more details here : https://www.postgresql.org/docs/current/app-pgdump.html - look for format. I will add a comment

services/pg-backup/docker-compose.yml
mrnicegyu11: OK this sounds good to me, it's better to fail than to have a weird behaviour.

services/pg-backup/docker-compose.yml
mrnicegyu11: OK I want to give it one last try to persuade you:
In the past, for the pg14 upgrade, I had to manually mess with the `.sql` files since there was a custom function defined for some relations in the DB that was wrong, but pg11 was fine with it. PG14 however would not accept it. From this experience, I personally am happy to accept 7 times bigger backup sizes, since anyway this is hosted as a service for us by O+P. So I dont see a problem with large space needs. We should delete old backups older than >1 year anyway every now and then ;)

Your thoughts?

Surfict: I agree, let's go for it.

Surfict: I added a policy to delete them after 1 year.

services/pg-backup/Makefile
Surfict: Done

PR #85, opened 2023-01-10
Divers fix to make pg-backup works on all deployments. ( :warning: ) 

Surfict:
- Readme
- Remove prune while deploying to allow two PG-backup in the same stack (Dalco staging and dalco production)
- Parametrization of the service name for osparc-public


mrnicegyu11:
( :warning: ): we need to check that the docker node label `pgbackup` is set on the machine that has the tank mounted

services/pg-backup/docker-compose.public.yml.j2
```diff
@@ -0,0 +1,5 @@
+version: '3.7'
+services:
+  {{ PREFIX_STACK_NAME }}_pg_backup:
+    volumes:
+    - /var/run/docker.sock:/var/run/docker.sock
```
mrnicegyu11, line 5: Why does this need to have the docker socket mounted for public?

mrnicegyu11 (commented):
I have one question, otherwise thanks a lot looks good

services/pg-backup/docker-compose.public.yml.j2
Surfict: It is from the precedent pg-backup. I will remove it. Thanks

mrnicegyu11 (approved):

PR #86, opened 2023-01-18
🐛 Fix: Dont capture service.testing.${MACHINE_FQDN} in maintenance page

mrnicegyu11:
This PR fixes a bug where routes matching `service.testing.${MACHINE_FQDN}` would be captured by and forwarded to the maintenance page. In order for devs to test deployments via the `testing.${MACHINE_FQDN}` route, the previously mentioned dy-service routes need to be forwarded to the associated services' containers, even when the maintenance page is up.

This PR fixes this by introducing a properly devised traefik-rule in the `common.Makefile`, which is built from the `DEPLOYMENT_FQDNS` env-var in the osparc-ops-deployment-configuration, and used by the maintenance page.


Surfict:
THanks !

Surfict (approved):

GitHK (approved):

pcrespov (commented):
👍

sanderegg (approved):
great!

PR #91, opened 2023-01-25
🐛 Fix bug in BucketCORS script

mrnicegyu11:
🐛 Fixes bug in BucketCORS script due to wrong variable capitalization

Bonus: 
- Adds `launch.bash` for script
- Pins pip reqs for script


scripts/s3BucketOptions/setBucketLifecyclePolicy.py
```diff
@@ -14,15 +14,15 @@ def main(
     destinationbucketaccess: str,
     destinationbucketsecret: str,
     destinationendpointurl: str,
-    noncurrentVersionExpirationDays: int,
+    noncurrentversionexpirationdays: int,
```
pcrespov, line 17: These variable names are really difficult to read as a single word: use ``snake_case`` for variables and ``CamelCase`` for classes. I would start considering re-formatting all the python code in this repo as in ospar-simcore.

Same for the filenames.

scripts/s3BucketOptions/setBucketLifecyclePolicy.py
```diff
@@ -41,6 +41,7 @@ def main(
         print("########")
     except botocore.exceptions.ClientError as e:
         if e.response["Error"]["Code"] == "NoSuchLifecycleConfiguration":
+            print("Lifecycle Settings before the change:")
```
pcrespov, line 44: Use loggers instead of prints to display info. You will have a chance in the future to capture structured log messages in the future

scripts/s3BucketOptions/launch.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#
+
+set -o nounset
+set -o pipefail
+IFS=$'\n\t'
+
+if [ -f ".venv/bin/activate" ] ; then
+	sleep 0
```
pcrespov, line 9: Q: why the sleep is necessary here ? why not just check that the file does not exists?

scripts/s3BucketOptions/setBucketLifecyclePolicy.py
mrnicegyu11: good point thanks for the input. feel free to open a ticket about it :) code formatting is imho out of the scope of this minor PR :)

scripts/s3BucketOptions/setBucketLifecyclePolicy.py
mrnicegyu11: good point thanks for the input. feel free to open a ticket about it :) code cleanup is imho out of the scope of this minor PR :)

scripts/s3BucketOptions/launch.bash
mrnicegyu11: It is only for readability, since if/else checks in bash are a bit tricky. `sleep 0` is an empty statement, it does nothing
This reads as:
`If the venv exists, do nothing, else, do this/that`

Surfict (approved):

scripts/s3BucketOptions/setBucketLifecyclePolicy.py
mrnicegyu11: good point thanks for the input. feel free to open a ticket about it :) code cleanup is imho out of the scope of this minor PR :)

sanderegg (commented):
:+1:

scripts/s3BucketOptions/setBucketLifecyclePolicy.py
pcrespov: https://github.com/ITISFoundation/osparc-ops-environments/issues/97

pcrespov (commented):
:+1:

PR #92, opened 2023-01-25
🐛 Fix error in call_make call

mrnicegyu11:
This was uncovered when trying to prepare an ops-PR going with simcore-pr https://github.com/ITISFoundation/osparc-simcore/pull/3787

It would prevent `make up-local` from running successfully.

Bonus:
- adds silent pushd call
- adds retry to deployment without deployment agent

### DIFF
--- a/scripts/deployments/deploy.sh
+++ b/scripts/deployments/deploy.sh
@@ -22,6 +22,11 @@ function call_make
     make --no-print-directory --directory "${1:-"Missing Directory"}" "${@:2:$#}"
 }
 
+# Silent pushd via https://stackoverflow.com/a/25288289
+pushd () {
+    command pushd "$@" > /dev/null
+}
+
 
 declare psed # fixes issue with not finding psed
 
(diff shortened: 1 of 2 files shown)


scripts/deployments/deploy.sh
```diff
@@ -147,7 +152,7 @@ log_info "Deploying osparc for $1 cluster on ${MACHINE_FQDN}, using credentials
 if [ "$disable_vcs_check" -eq 1 ]; then
     log_info "Asserting that there are no uncommited changes in the config-files docker-compose-deploy and .env ..."
     pushd "${repo_basedir}"/services/simcore;
-    call_make "${repo_basedir}"/services/simcore compose-"$1" > /dev/null
+    call_make . compose-"$1" > /dev/null
```
pcrespov, line 155: no quotes here? i.e. ``"."``
do you have shellcheck extension? If not, hit is highly recommended : timonwong.shellcheck

GitHK (approved):

Surfict (approved):

PR #105, opened 2023-02-15
Fix pg_backup starts

Surfict:
When pg-backup is started before the Postgres container, it will fail with this error :
image
The container itself doesn't fail and still runs, but doesn't do any backup. To solve this, we need to ensure that pgsql container is started and ready to accept connections before starting pg-backup.

Fix #443 ops issue.

### DIFF
--- a/scripts/deployments/deploy.sh
+++ b/scripts/deployments/deploy.sh
@@ -254,17 +254,6 @@ if [ "$start_opsstack" -eq 0 ]; then
     call_make "." up-"$stack_target"
     popd
 
-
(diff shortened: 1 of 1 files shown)


mrnicegyu11 (approved):
Thanks a lot for this!

Quick question: What happens when the postgres container is restarted (for example when releasing new code)

Surfict:
> Thanks a lot for this!
> 
> Quick question: What happens when the postgres container is restarted (for example when releasing new code)

I didn't test it but from my comprehension of how pg-backup works, it still works. It will check the container at startup and fails without doing backup if it doesn't find it. If it can't do a backup later, it will try again at the next planned window. 

I guess if I am wrong and it prevents if from functioning, we will see it in the tests, but so far I have never seen it after months of functioning so I believe it works.

PR #115, opened 2023-03-14
change agent deploy policy

GitHK:
Since the agent registers RPC calls on startup and these cannot overlap, it is required that the service is fully stopped before a new instance is started.

@mrnicegyu11 I hope this is the correct place to change it

- required by https://github.com/ITISFoundation/osparc-simcore/pull/3941

### DIFF
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -66,7 +66,7 @@ services:
           - node.labels.dynamicsidecar==true
       update_config:
         parallelism: 2
-        order: start-first
+        order: stop-first
         failure_action: rollback
         delay: 10s
       rollback_config:


mrnicegyu11:
thanks a lot for this! It is the correct place to change it! I am not an expert for the particular change you made, i.e. for the restart-policies, maybe sylvain can better assess if this change brings what you need.

Supernice, thanks again!
Please somehow reference/tag this PR in your associated simcore PR so it is not forgotten that they go together :)

mrnicegyu11 (approved):

GitHK:
@mrnicegyu11 it is linked back with instructions that it is required when deploying to be in the target deployment

PR #118, opened 2023-03-16
Add Graylog Alert For Unhandled Errors And Exceptions - running every 24h

mrnicegyu11:
Dear backend, I think we should trace and fix unhandled exceptions. This alert might help to remind us that they exist.

Up for debate.


mrnicegyu11:
I propose we slowly move to try to have "things not to worry about" to not print `Traceback` or `Error`, then this check might help us really find unexpected errors we might skip otherwise

GitHK (commented):
@mrnicegyu11 How will this be displayed where will this data go?

mrnicegyu11:
> @mrnicegyu11 How will this be displayed where will this data go?
This would go into the oSparc Batman Mattermost channel as usual. No info about the alerts would be displayed there I guess. But we can adjust this, what do you have in mind?

mrnicegyu11:
merging this before it becomes forgotten, I prefer reverting in case it just creates noise.

PR #123, opened 2023-03-22
🔊 Alerting/add grafana alert for logspout swapping

mrnicegyu11:
logspout using more memory than allowed and swapping can lead to heavy IO and sluggish machine states. This alert checks if logspout is swapping. If logspout swaps, even higher memory allocation for this guy is needed.

In the future, we might move to dual logging to avoid this (see: https://git.speag.com/oSparc/osparc-ops-environments/-/issues/418 and https://git.speag.com/oSparc/osparc-infra/-/issues/incident/20 )

*Bonus:*
🐛 Fixes Bug: More than one grafana alert rule can now be imported


services/monitoring/grafana/scripts/export.py
```diff
@@ -106,15 +106,25 @@ def main(foldername: str = ""):
...
-            alert["rules"][0]["grafana_alert"].pop("uid", None)
-            # Remove orgId
-            alert["rules"][0]["grafana_alert"].pop("orgId", None)
-            # Remove id
-            alert["rules"][0]["grafana_alert"].pop("id", None)
-            # Remove id
-            alert["rules"][0]["grafana_alert"].pop("namespace_id", None)
-            # Remove id
-            alert["rules"][0]["grafana_alert"].pop("namespace_uid", None)
+            for ruleIter in range(len(alert["rules"])):
```
pcrespov, line 109: Some TIPS:

1. should start using same coding conventions as other repos (it is far easier to read when we have all the same style!)
2. this loop can simply be done by
```python
for rule in alert['rules']:
    rule["grafana_alert"].pop("uid", None)
   ...

```

services/graylog/docker-compose.yml
```diff
@@ -108,10 +108,10 @@ services:
       resources:
         limits:
           cpus: "0.20"
-          memory: 256M
+          memory: 512M
```
pcrespov, line 111: NOTE: at some point we should revisit a common strategy to review how and what we log! 
E.g. this issue since 2021 https://github.com/ITISFoundation/osparc-simcore/issues/2157

services/monitoring/grafana/scripts/export.py
mrnicegyu11: changed this thx

PR #124, opened 2023-03-22
⬆️ Upgrade: docker compose v1 deprecation

mrnicegyu11:
- ⬆️ Adds `docker-compose-config.bash` from osparc-simcore as suggested in https://github.com/ITISFoundation/osparc-ops-environments/pull/120#discussion_r1143147613
- ⬆️ Implements this everywhere

*Tested on local deployment*


pcrespov (commented):
ack

PR #127, opened 2023-03-24
🚧🚧🚧 Add traefik routing for invitation services

mrnicegyu11:
This adds traefik routing rules for `invitations.osparc.io` and forwards it to the invitation service. It only works with the `MACHINE_FQDN` for now (so, to make this clear, only `invitations.osparc.io` works, not `invitations.s4l-lite.io`). This was tested to work on my local deployment. This should ideally move fast to prod.

🚧🚧🚧: 
 - This deprecates https://git.speag.com/oSparc/invitations and the machines and the repo can be removed.
 - This goes with https://git.speag.com/oSparc/e2e-ops/-/merge_requests/53


pcrespov (commented):
Question: would it be possible to keep a prefixed host name (i think the right term is subdomain)  i.e. ``invitations.*``  instead of adding as part of the path ??

This way we do not need to notify third parties to change the routing

mrnicegyu11:
It is possible but this complicates the routing a bit, since prefixes (subdomains) are managed by the ops-traefik. If one uses `/invitations`, the routing can be fully handled on the simcore traefik. However, as I notice now (thanks for poiting me in this direction), the invitation service is then affected by the maintenance page (i.e. unavailable if the maintenance page is up). If this is not desirable (and currently there is no urgency), both issues can potentially be addressed. 

let me know again if you think :)

GitHK (approved):

matusdrobuliak66 (approved):

mrnicegyu11:
Caught by @Surfict: Entrypoint should be `https` and thus needs `myresolver`

mrnicegyu11:
Todo: Add new subroute to certificates test

sanderegg (commented):
should I review this or you are still checking for @pcrespov questions?

sanderegg (approved):

pcrespov (approved):
Thx!
Make sure to write a big deprecation/retire note in https://git.speag.com/oSparc/invitations  pointing to the PR

PR #130, opened 2023-03-29
♻️ Alert/add hanging sidecar alert + improve grace periods

mrnicegyu11:
This PR
- Adds a new alert for hanging sidecars that require human intervention
- Changes alerting grace periods on grafana alerts to avoid spam
- Changes how alert for pending docker services work: it now asserts that at least at one instance (once) in the last 2h there were no pending docker services. This removes autoscaling related false positives on aws-prod

### DIFF
--- a/services/graylog/scripts/alerts.template.yaml
+++ b/services/graylog/scripts/alerts.template.yaml
@@ -271,12 +271,12 @@
     grace_period_ms: 0
     backlog_size:
   alert: true
-- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
-  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
+- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected."
+  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Human intervention required. Please investigate."
   priority: 3
   config:
     query: >
-      "unexpectedly failed [OEC" AND container_name:/.*director-v2.*/
+      "not contact dynamic-sidecar to save service state or output ports" AND container_name:/.*director-v2.*/
     query_parameters: []
     search_within_ms: 3600000
     execute_every_ms: 3600000
@@ -305,7 +305,6 @@
   key_spec:
     - source
     - container_name
-    - full_message
   notification_settings:
     grace_period_ms: 0
     backlog_size:
--- a/services/monitoring/grafana/provisioning/allDeployments/alerts/osparc_alerts.json
+++ b/services/monitoring/grafana/provisioning/allDeployments/alerts/osparc_alerts.json
@@ -4,7 +4,7 @@
   "rules": [
     {
       "expr": "",
-      "for": "3m",
+      "for": "10m",
       "grafana_alert": {
         "condition": "B",
         "data": [
@@ -83,7 +83,7 @@
     },
     {
       "expr": "",
-      "for": "5m",
+      "for": "2h",
       "grafana_alert": {
         "condition": "C",
         "data": [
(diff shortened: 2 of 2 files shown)


services/graylog/scripts/alerts.template.yaml
```diff
@@ -271,12 +271,12 @@
...
   alert: true
-- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
-  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
+- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected."
+  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Human intervention required. Please investigate."
   priority: 3
   config:
     query: >
-      "unexpectedly failed [OEC" AND container_name:/.*director-v2.*/
+      "not contact dynamic-sidecar to save service state or output ports" AND container_name:/.*director-v2.*/
```
pcrespov, line 279: NOTE that it is very easy get error message out of sync (we are constantly changing those to make them more "human readable"). I would use instead some sort of identifier code, similar  to the ``OEC`` that you have just removed.

Some suggestions:
1. define some unique "graylog markers" in osparc-simcore 
2. append them in logs from errors you want to detect 

If you like we can create a mechanism that does that automatically. @matusdrobuliak66 might be working in improving the loggers so we could add it. Just keep it in mind.

services/graylog/scripts/alerts.template.yaml
```diff
@@ -271,12 +271,12 @@
...
     backlog_size:
   alert: true
-- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
-  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Please investigate."
+- title: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected."
+  description: "${MACHINE_FQDN}: Potential hanging dy-sidecar service detected. Human intervention required. Please investigate."
   priority: 3
   config:
     query: >
-      "unexpectedly failed [OEC" AND container_name:/.*director-v2.*/
```
pcrespov, line 279: Q: why not to keep the OEC-like errors?

mrnicegyu11: Because there were >150 (for all 6 deployments summed) in 24h, so this is too spammy for an "escalation to human" at this point :(

Once we have few enough OECs that one popping up can be categorized as "unexpected", we should indeed have the OEC alert again ;)

services/graylog/scripts/alerts.template.yaml
mrnicegyu11: You are correct, and a system for synchronizing this would be nice :) But conceptually this amounts to synchronizingv the ops-repo and the simcore repo, and this synchronization will have to work across multiple deployments and code-versions (releases). So to have this robust this is a big task. Ideally, it would be a system as you propose though

sanderegg: small suggestion:
- why not define a unique ID (like this OEC I think is), and create a OR case. This way you ensure you still get the text version (which has all the issues related here by @pcrespov) and we slowly move to the ID based one. This way I think you would cover all the deployments and in a few releases you can remove the text-based one?

sanderegg (approved):
Thanks for taking care of this. I would be interested into better understanding how the
- graylog alerting works (I believe this ends in Mattermost)
- the grafana alerting works? Does this also end up in mattermost or is there a dashboard?

services/graylog/scripts/alerts.template.yaml
GitHK: So I like this unique ID to tag to these errors that require monitoring so that messages can be caught reliably in Graylog.

Also not to be a party buster, I found out that this message is also not that reliable. It can get triggered multiple times due to retries and in the end the sidecar is removed successfully. 

My [upcoming PR](https://github.com/ITISFoundation/osparc-simcore/pull/4040) will provide one that only shows one when a sidecar is detected as waiting for manual user intervention.

GitHK: FYI: updated PR to reflect changes from https://github.com/ITISFoundation/osparc-simcore/pull/4040/files#diff-580b3ba278d2b6f71ef7fa38a284a0b5a5ef7cbf506f8ba7a77990cb22238da0R40-R42

GitHK (approved):
👍 fine with me now!

PR #131, opened 2023-03-29
🗑️Remove mailserver automatic starting from aws

mrnicegyu11:
We now use a AWS SES based mail-delivery, the docker-mailserver is only used for the local deployment now.


Surfict (approved):

mrnicegyu11:
> Q: should we keep the mail config/server around ... in the spirit of having always a non-aws alternative (as with pg, minio etc).

Yes it will stay around at least for the local deployment of the ops-stack, which is used for debugging :)

PR #134, opened 2023-04-11
Update S4L lite admin user dashboard

mrnicegyu11:
- Filters out puppeteer users (requires simcore release v1.52.0)
- Fixes bug in CPU usage panel

looks like so:
image


odeimaiz (approved):
👍

PR #148, opened 2023-05-02
🚧🐛 Fix deployment agent config on clusters with multiple deployments

mrnicegyu11:
On clusters running multiple simcore stacks, such as osparc-dalco, a bug was discovered. All deployment-agents would run using the same config-files (accessed via docker configs), and the simcore stacks might get into a weird state, neither running fully staging or production configuration.

This PR addresses this by inserting the STACK_PREFIX into the name of the docker configs using j2 templates, thereby enforcing a single unique docker config per simcore stack.

🚧🚧🚧:
Once this propagates through the deployments, the docker stacks of the deployment agents need to be re-created.

### DIFF
--- a/services/deployment-agent/.gitignore
+++ b/services/deployment-agent/.gitignore
@@ -1,1 +1,2 @@
 deployment_config.yml
+docker-compose.yml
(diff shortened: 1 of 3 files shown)


Surfict (approved):
Nice work, thanks !

sanderegg (approved):

PR #149, opened 2023-05-02
Fix bad configuration for dns challenge resolver

Surfict:
This PR : 

- Fix a bad configuration for DNS challenges that are done by Traefik. Let's encrypt servers couldn't solve properly the challenge because the variable used to indicate which DNS server to look at while doing so was not properly set up.


mrnicegyu11 (approved):
Thanks!

PR #150, opened 2023-05-04
Prometheus swarm target bug

Surfict:
This PR : 

Fixes a bug from Prometheus targets configuration. More details [here](https://github.com/prometheus/prometheus/issues/11060#issuecomment-1195278301). This bug was impacting aws-production deploy (not able to scrape the manager of the swarm)


mrnicegyu11 (approved):
Good find, it seems like we have yet another "cornercase" where swarm is not working easily with other tools. Anyway, glad it is taken care of :) thx

PR #152, opened 2023-05-08
Deployment script - pg-backup improvments

Surfict:
This PRs : 

- Makes a better check about when to start pg-backup in the cases we have several deploys in the same cluster
- Fixes issue #543 (check was taking postgres exporters as the pgsql container)


scripts/deployments/deploy.sh
```diff
@@ -292,26 +292,40 @@ if [ "$start_simcore" -eq 0 ]; then
...
     # -------------------------------- BACKUP PG -------------------------------
     # PG-backup has to wait for postgres container to be started and ready before starting, or it will fail.
-    # Wait for potsgres container to start
-    log_info "Before starting PG-backup, we ensure that postgres is started and ready to accept connections."
-    until docker ps -a --format '{{.Names}}' | grep -q "postgres"; do
-      log_info "Postgres didn't start yet. Waiting for 5 seconds..."
-      sleep 5
+    # Wait for all potsgres container to start
+    log_info "Before starting PG-backup, we ensure that all postgres are started and ready to accept connections."
+    until [[ $(docker ps --format "{{.Names}}" | grep "simcore_.*_postgres" | wc -l) -eq $deploys_count ]]; do
```
mrnicegyu11, line 311: I think in all these calls to `docker ps` you assume that postgres runs on the same node as the one that the scripts runs on. This does not need to be the case. Maybe it would be better to check the status of the docker services relating to postgres?

scripts/deployments/deploy.sh
```diff
@@ -292,26 +292,40 @@ if [ "$start_simcore" -eq 0 ]; then
     fi
 fi
 # shellcheck disable=2235
-if [ "$start_opsstack" -eq 0 ] && ([ "$stack_target" = "dalco" ] || [ "$stack_target" = "master" ] || [ "$stack_target" = "public" ]); then
+
+# We pause 2 mns to be sure the deployment agent is launched
+sleep 120
```
mrnicegyu11, line 297: Could we live without this? A 2 min sleep without any log-message telling the operator that the script will now sleep will really look like there is some bug. I prefer not to have any sleeps outside retrying-functionality (the while loops you made below are retrying, so there it is fine)

mrnicegyu11 (changes requested):
Some feedback, thx

scripts/deployments/deploy.sh
Surfict: Done

scripts/deployments/deploy.sh
mrnicegyu11: But the script `scripts/deployments/deploy.sh` does not necessarily run on the node that has the label `postgres=true`. So docker ps here wont work

Surfict: Fixed using docker services in place of docker ps.

mrnicegyu11 (approved):
super, thanks a lot!

PR #153, opened 2023-05-10
Add/deployment agent for graylog

mrnicegyu11:
PoC to add GitOps to ops-stack tools *without refactoring the deployment agent*. 

I set out to spend some hours to see whether there is a low hanging fruit way of automatically pushing the state from the github code repos to deployments (this is called gitops) for graylog. We already have a gitops workflow in place for the simcore Stack, using our custom deployment Agent. 

Benefits of gitops:
Ideally, backend developers should have a smooth way to add and remove log based alerts for critical errors using graylog. The current pain point here is that graylog ( on All deployments) does not have gitops, and thus any changes to the graylog alerts need to be propagated manually by the ops team. A gitops workflow would unlock backend devs, so they may potentially manage their own alerts using PRs and git. 

Background: the deployment agent uses (for historical reasons) the portainer API to Update a docker swarm stack. The docker-compose.yml can be generated automatically by running a user provided bash Script Inside the monitored git repositories. Using the (potentially bash Script generated) docker-compose.yml, the deployment Agent also watches for changes in the images of the provided docker Containers, for example on dockerhub. 

Why simply copying the gitops workflow of osparc-simcore didnt work:
I tried to replicate the usual simcore gitops workflow and it was not viable for graylog. The reason is that graylog uses docker configs, which cannot be easily provisioned via the portainer REST API, since portainer expects to be able to access the source files from its (portainer's) local Filesystem. Moving away from docker configs entails limitations in docker swarm Container placement. If possible, we should continue to use configs. 

Alternative approach ideally:
The deployment Agent could be refactored to use aiodocker instead of portainer rest calls. Currently, the portainer interface of the deployment Agent is tightly unit tested and reliable. Making a aiodocker interface Just as robust will be some work. 
As yet another Option, we could simply run a user defined bash command upon a change in the git repos (ie docker Stack deploy), skipping any subsequent steps. This would make the deployment Agent more universal, it is then a General Tool to execute bash commands upon changes in git repos. However, this will break the current workflow of the docker registry being monitored as well, since watching a list of docker images depends on the existance of a valid docker-compose.yml file from where to get the docker image names. We would have to remove the docker image watcher functionality in this case. 

Hacky approach: PoC in this PR 
If we are willing to try hacky stuff with The current deployment Agent Code, we can abuse the User defined bash command that the deployment Agent runs to have it execute 'docker Stack deploy' as well. Afterwards we give the deployment Agent an Empty docker-compose.yml file, so the deployment Agent will Not start Stacks using the portainer API afterwards. This works, but the deployment Agent will then actually always execute this command in every polling interval (60 sec), no matter if there have been actual file content changes in the git repo. Works for graylog, but is arguably not ideal. Like I said, a hacky approach. 

Thanks for reading so far, I please require ideas and input if and how to proceed with this. What would you say is a useful way forward? 

Thanks! 

----------

Depends on experimental changes in the deployment-agent https://github.com/ITISFoundation/osparc-deployment-agent/pull/206

ToDo:
- Add bash retires for docker stack deploy etc.

- write test


sanderegg:
@mrnicegyu11 
After reading your description and check a bit the code, here are a few things I would do (of course up to discussion):
- remove the dependency to Portainer:
   1. If I would re-code the deployment-agent today I would not use the portainer API at all but only the docker API
   2. I think the changes you need to do are not so difficult and I'm happy to assist. we can check the to have the necessary tests in place so you can go for aiodocker easily.
   3. The only usefulness of Portainer as I see it are the agents, that allow to list containers over the swarm, but this is not something that the deployment-agent needs
   4. This would make the whole process cleaner, and remove the need to hack
- the docker config:
  1. How do you update the docker configs? As far as I know the only way is to delete/re-create the configuration right?
  2. if yes, how can you update the stack without putting it down? or you create new configs and override the old ones?

mrnicegyu11:
deployment agent has been removed

PR #179, opened 2023-06-19
Adds second prometheus

mrnicegyu11:
Adds a prometheus specifically for tracking cAdvisor metrics for 30 days.

This is functional, and doesn't interfere with the current "regular" Prometheus.

The prometheus is exposed at `monitoring.osparc.local/prometheuscadvisor`

### DIFF
--- a/services/monitoring/docker-compose.aws.yml
+++ b/services/monitoring/docker-compose.aws.yml
@@ -1,6 +1,6 @@
 version: "3.7"
 services:
-  prometheus:
+  prometheus-catchall:
     dns: 8.8.8.8
     deploy:
       placement:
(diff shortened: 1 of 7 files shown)


matusdrobuliak66:
Q: would it be possible to setup that these two prometheus guys live on different nodes?

mrnicegyu11:
> Q: would it be possible to setup that these two prometheus guys live on different nodes?

yes, would you want that? If so, only on aws-prod, right?

sanderegg (approved):
thanks a lot. can you also update the credentials page?

matusdrobuliak66:
> > Q: would it be possible to setup that these two prometheus guys live on different nodes?
> 
> yes, would you want that? If so, only on aws-prod, right?

sorry for the late answer! yes, i guess it would make it more robust :+1: yes aws-prod is the most important

PR #180, opened 2023-06-21
Bump and pin versions

mrnicegyu11:
Bump and pin versions

Tested locally, so far no problems.


matusdrobuliak66 (approved):

sanderegg (approved):
I think that is a good idea.
Nevertheless why don't you put that in the osparc-ops-environement-versions ?

PR #184, opened 2023-06-23
🐛 Fix 302 Redirect bug in cadvisor prometheus

mrnicegyu11:
The line ` - "--web.external-url=https://${MONITORING_DOMAIN}/prometheus-cadvisor/"` was wrong.

Bonus:
Fix ambiguous naming in the two Prometheus services.


matusdrobuliak66 (approved):
Thanks!

PR #303, opened 2023-08-08
Remove MAIL service

YuryHrytsuk:
## What do changes do
These changes remove MAIL services from the repository. Since we no longer rely on self-hosted MAIL service, we don't need it anymore. Furthermore, it causes errors while deploying locally and causes certificates generations (`./certificates`) to fail due to missing mail-related variable

## How to test
Deploy OPS stack locally and ensure that MAIL services is not deployed and all services are up and running


mrnicegyu11 (approved):
superb, thanks! please merge in

PR #305, opened 2023-08-08
Change dependency order in deployment-agent Makefile

YuryHrytsuk:
## What do these changes do
These changes fix a minor bug in a `Makefile` of deployment-agent service. The dependency order was wrong causing error on first launch (`up-<target>`)

## How to test
Deploy OPS stack from scratch and ensure that deployment agent launched successfully from the first try, i.e. no errors occurred


mrnicegyu11 (approved):
perfect thank you

PR #306, opened 2023-08-08
✨ Add docker node resources statistics script

mrnicegyu11:
Adds a script as asked for by @mguidon which inspects the available CPU and Memory resources on the swarm machines and compares them to the limits of the docker services running on those machines.

This script needs to be run on a swarm manager node.

Currently the output is ambiguous because we dont have strict limits and reservations set on all services.

### Bonus:
- New `itisfoundation/ci-provisioned-ubuntu:0.3.0` docker image with the docker-cli installed, able to run docker in docker.


scripts/dockerNodeStats/launch.bash
```diff
@@ -0,0 +1,19 @@
...
+IFS=$'\n\t'
+
+if [ -f ".venv/bin/activate" ] ; then
+	sleep 0
+else
+    echo "Installing and creating venv.."
+    pip3 install virtualenv >/dev/null 2>&1
+    virtualenv -p python3 .venv >/dev/null 2>&1
+fi
+set -o allexport
```
sanderegg, line 15: question: why don't you run these things in a docker container? then you would not contaminate the machines where you run this. There are plenty of examples in osparc-simcore/scripts

scripts/dockerNodeStats/main.py
```diff
@@ -0,0 +1,197 @@
+import csv
+import json
+import logging
+import os
+import subprocess
+from collections import defaultdict
+
+import tqdm
+
```
sanderegg, line 9: I would propose that we start trying that library:
https://github.com/amoffat/sh

6.7Kstars... seems pretty cool for wrapping these subprocesses... found it through some articles.

sanderegg (approved):

scripts/dockerNodeStats/launch.bash
mguidon: I agree.

scripts/dockerNodeStats/main.py
```diff
@@ -0,0 +1,197 @@
...
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+
+# Add handler to logger
+logger.addHandler(handler)
+logger.propagate = False
+
+# Use logger
+logger.info("Starting...")
+service_ids: list[str] = (
```
YuryHrytsuk, line 30: I would avoid running everything on global level since this is a bad practice. A good practice is to have `if __name__ == "__main__": ...` and further function calls.

However, in this particular case it is clearly not a must.

scripts/dockerNodeStats/main.py
YuryHrytsuk: I would avoid executing bash in Python when it is possible

YuryHrytsuk (approved):
The code is clear and does its job :+1:  I tested in on locally and on master. Kudos :tada:

scripts/dockerNodeStats/launch.bash
mrnicegyu11: done

mrnicegyu11:
> Thanks, looking forward to test this. One minor thing: Can we also get hold of generic resources? (e.g. VRAM?)

done

scripts/dockerNodeStats/main.py
mrnicegyu11: I intriduced `sh`. @YuryHrytsuk you are absolutely right, this was based on some script by an AI-Bot, who already used a lot of shell-calls, and was provided to me by @mguidon . If we code something properly ourselves we should indeed avoid console calls and use packages, modules and APIs :)

sanderegg (approved):
nice. and this sh library looks pretty.
thanks for the changes.

PR #312, opened 2023-08-15
✨ Add deployment labels for clusters_keeper microservices

mrnicegyu11:
- Adds restart_config and resource limits for simcore `clusters_keeper` microservice (see https://github.com/ITISFoundation/osparc-simcore/pull/4591  )

## Bonus:
- Update pre-commit hooks
- Update pre-commit hooks' used python version to 3.10
- Removes deprecated file comment about logspout (which we dont use anymore)

### DIFF
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -1,9 +1,44 @@
-# osparc-simcore stack (framework stack)
-# NOTES :
-# LOGSPOUT_MULTILINE is used to tell to Logspout, which is used by GrayLog, to handle multiline support of the logs of the concerned container. Please add with every new osparc-simcore service.
-# See this issue to have more informations : https://github.com/ITISFoundation/osparc-ops/issues/40
 version: "3.8"
 services:
+  clusters-keeper:
+    hostname: "{{.Node.Hostname}}-{{.Service.Name}}"
+    environment:
+      - LOG_LEVEL=${LOG_LEVEL:-WARNING}
+      - LOG_FORMAT_LOCAL_DEV_ENABLED=${LOG_FORMAT_LOCAL_DEV_ENABLED}
+      - CLUSTERS_KEEPER_LOGLEVEL=${LOG_LEVEL:-INFO}
+      - REDIS_HOST=${REDIS_HOST}
+      - REDIS_PORT=${REDIS_PORT}
+      - RABBIT_HOST=${RABBIT_HOST}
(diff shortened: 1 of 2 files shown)


services/simcore/docker-compose.yml
```diff
@@ -1,9 +1,44 @@
-# osparc-simcore stack (framework stack)
-# NOTES :
-# LOGSPOUT_MULTILINE is used to tell to Logspout, which is used by GrayLog, to handle multiline support of the logs of the concerned container. Please add with every new osparc-simcore service.
-# See this issue to have more informations : https://github.com/ITISFoundation/osparc-ops/issues/40
 version: "3.8"
 services:
+  clusters-keeper:
+    hostname: "{{.Node.Hostname}}-{{.Service.Name}}"
```
sanderegg, line 4: in osparc-simcore we define it as: ```hostname: "{{.Node.Hostname}}-{{.Service.Name}}-{{.Task.Slot}}"```
The task.Slot copes with the fact that in case of scaling up the service and more than 1 replica ends up on the same machine that they still have different names.
This only does not work in case of global service. I would propose to just remove that line.

services/simcore/docker-compose.yml
```diff
@@ -1,9 +1,44 @@
...
+      - LOG_LEVEL=${LOG_LEVEL:-WARNING}
+      - LOG_FORMAT_LOCAL_DEV_ENABLED=${LOG_FORMAT_LOCAL_DEV_ENABLED}
+      - CLUSTERS_KEEPER_LOGLEVEL=${LOG_LEVEL:-INFO}
+      - REDIS_HOST=${REDIS_HOST}
+      - REDIS_PORT=${REDIS_PORT}
+      - RABBIT_HOST=${RABBIT_HOST}
+      - RABBIT_PASSWORD=${RABBIT_PASSWORD}
+      - RABBIT_PORT=${RABBIT_PORT}
+      - RABBIT_SECURE=${RABBIT_SECURE}
+      - RABBIT_USER=${RABBIT_USER}
```
sanderegg, line 15: this is good. As long at it has no EC2_ACCESS variables defined it will do nothing.
there will be a PR later when it becomes usable.

sanderegg (approved):
Please check the hostname part in docker-compose. this is not unique

services/simcore/docker-compose.yml
mrnicegyu11: thats correct I guess, thanks for catching it, will remove the line

PR #341, opened 2023-09-07
Fix aws stag registry routing

YuryHrytsuk:


YuryHrytsuk:
We should not hardcode things!

PR #361, opened 2023-09-20
Minor fixes

YuryHrytsuk:
1. Update `make down` in deployment agent. From now on, it will NOT remove simcore stack
2. Disable debug mode by default in `stratup_script` of deployment_agent
3. Execute `check stack yml is generated` the last in pre-commit hooks


mrnicegyu11 (approved):
thanks a lot!

PR #372, opened 2023-09-25
Disable payments service on aws & dalco

YuryHrytsuk:
Payments service is not yet ready that is why for safety we disable it on aws, dalco and tip deployments. Thus, it will be only running on master as of now.


pcrespov (approved):
Thx! 
But we would have to re-enable it in staging by the end of this week.

PR #396, opened 2023-10-18
Remove letsencrypt dns challenge propagation delay

YuryHrytsuk:
This removes the hardcoded DNS challenge propagation delay which I find unnecessary. It doesn't solve the certificate issue we have.

### DIFF
--- a/services/traefik/docker-compose.aws.yml
+++ b/services/traefik/docker-compose.aws.yml
@@ -24,7 +24,7 @@ services:
       - "--providers.docker.constraints=!LabelRegex(`io.simcore.zone`, `.+`)"
       - "--entryPoints.https.forwardedHeaders.insecure"
       - "--certificatesresolvers.myresolver.acme.dnschallenge.provider=route53"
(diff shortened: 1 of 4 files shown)


matusdrobuliak66 (approved):

services/traefik/docker-compose.aws.yml
```diff
@@ -24,7 +24,7 @@ services:
       - "--providers.docker.constraints=!LabelRegex(`io.simcore.zone`, `.+`)"
       - "--entryPoints.https.forwardedHeaders.insecure"
       - "--certificatesresolvers.myresolver.acme.dnschallenge.provider=route53"
-      - "--certificatesresolvers.myresolver.acme.dnschallenge.delaybeforecheck=120"
+      # - "--certificatesresolvers.myresolver.acme.dnschallenge.delaybeforecheck=120"
```
pcrespov, line 27: I would advice against commenting code/configs. 
Rather delete and use version control to keep track of changes.

YuryHrytsuk: This is a good remark. I have deleted the commented lines

PR #401, opened 2023-10-20
Hide invitations swagger behind auth

YuryHrytsuk:
This PR does the following:
* simplifies (removes redundant settings)
* fixes swagger route (it should be behind basic auth)


mrnicegyu11 (approved):

pcrespov (commented):
thx

PR #402, opened 2023-10-24
Update transmitted ENV variables to clusters-keeper

sanderegg:
clusters-keeper ENV variables are now passed through in osparc-simcore repository by default.
Therefore no special treatment needed anymore in the ops-environment repository.

clusters-keeper is:
replicated 0 times everywhere, but on AWS-based deployments

### related simcore PR:
- PR4819 [https://github.com/ITISFoundation/osparc-simcore/pull/4899]


mrnicegyu11 (approved):
thanks a lot

PR #405, opened 2023-10-26
Add prepare simcore stack

YuryHrytsuk:


mrnicegyu11 (approved):
yes this is where the file should live actually, very good thanks!

Issue #426, opened 2023-11-10
Graylog fails to parse timestamp


YuryHrytsuk:
### Context
AWS STAG Graylog shows 70,000 of the following logs
```bash
ElasticsearchException[Elasticsearch exception [type=mapper_parsing_exception, reason=failed to parse field [log_timestamp] of type [date] in document with id '1ab0d7a3-7f5c-11ee-a560-0242ac120017'. Preview of field's value: '2023-11-10 00:00:05,658']]; nested: ElasticsearchException[Elasticsearch exception [type=illegal_argument_exception, reason=failed to parse date field [2023-11-10 00:00:05,658] with format [strict_date_optional_time||epoch_millis]]]; nested: ElasticsearchException[Elasticsearch exception [type=date_time_parse_exception, reason=Failed to parse with all enclosed parsers]];
```

YuryHrytsuk:
The error above must be caused by invalid format of date string. It should have `T` instead of ` ` as a separator between date and time. The documentation of [strict_date_optional_time](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#strict-date-time) attached below:
```
strict_date_optional_time: A generic ISO datetime parser, where the date must include the year at a minimum, and the time (separated by T)
    ```

YuryHrytsuk:
I can see that our graylog `indices` have  different type for `log_timestamp`. This behaviour is well described in this comment: 
https://community.graylog.org/t/elasticsearch-mapping-changed-from-keyword-to-long/28071/2

YuryHrytsuk:
This issue occurs randomly and disappears after index rotation (and probably new index has a proper `log_timestamp` type / format). I can see the same in AWS PROD. I conclude that this is a general problem and should be fixed by having a stricter configuration of indices (e.g. predefined index template).

YuryHrytsuk:
Well, the `type` of `log_timestamp` field should be set based on the first document written to the index. If it is `traefik` log then the type will be `date`, if `simcore` service logs then `keyword` since it is not valid date from `ES` default perspective. So, in case when `traefik` log arrives first and sets type to `date` all subsequent logs of `simcore` will not be parsed and fail.

YuryHrytsuk:
Fixed

PR #309, opened 2023-08-14
Extract traefik logs graylog & fix graylog configure

YuryHrytsuk:
## What do these changes do
1. Introduce/fix updating content packs
2. Add Traefik graylog extractor
3. Fix `configure` script for Graylog (https://github.com/ITISFoundation/osparc-ops-environments/issues/308)

## Related issues
Add extractor: https://github.com/ITISFoundation/osparc-ops-environments/issues/293
Fixes bug: https://github.com/ITISFoundation/osparc-ops-environments/issues/308

## How to test
1. Add a new extract to the content pack in `./data/contentpacks` directory
2. Update (bump/increase) revision number (key: `rev`) in the content pack file
3. Run `make configure` and check that the changes were propagated (system/inputs -> content packs & system/inputs -> manage extractors (at standard GELF UDP input)
4. Run `make configure` another time and make sure no changes were made (idempotence)

PR #430, opened 2023-11-13
BUGFIX:  Graylog cannot parse log_timestamp

YuryHrytsuk:
### Problem

This PR fixes the bug described in https://github.com/ITISFoundation/osparc-ops-environments/issues/426. In short, `simcore` services logs and `traefik` logs contain timestamp of different format which leads to occasional failure to parse any simcore service logs. 

### Solution

By stopping capturing timestamps generated by `traefik`, we make sure that all `log_timestamp` values have the same format

### Related PR
* https://github.com/ITISFoundation/osparc-ops-environments/issues/426
* https://github.com/ITISFoundation/osparc-ops-environments/pull/309

### For history and future usage

There are multiple solutions to this problem. We can create custom index mapping and tell explicitly `ES` which type should have `log_timestamp` but there is no easy solution to access `ES API` since `ES` is not reachable from the outside.

Other solution is to introduce `T` separator in all `log_timestamps` generated by `simcore` services (currently ` ` space is used as a separator which `ES date` type do no understand by default).


matusdrobuliak66 (approved):
Thanks! this one was nasty :)

mrnicegyu11 (approved):
THX

PR #438, opened 2023-11-15
Autoscaling: remove all overriden ENVs

sanderegg:
All these ENVs are defined in osparc-simcore

see https://github.com/ITISFoundation/osparc-simcore/pull/5033


YuryHrytsuk (approved):
Good direction. Thank you!

Issue #415, opened 2023-11-06
Show logs in `pg-backup` container


YuryHrytsuk:
### Actual behavior
No logs in `pg-backup` container are shown. Here are the only logs we get
```bash
Start script running with these environment options
DUMPPREFIX=PG_osparc-staging.speag.com
PG_CONN_PARAMETERS='-h staging_postgres -p 5432 -U postgres'
```

### Desired behavior
`pg-backup` container show logs about when it creates a backup, where it saves it and report any errors that happen.

### Objective
Sometimes we are missing backups and we do not get any insight on what happened from `pg-backup` container's logs. So, it is absolutely opaque and if it doesn't work we can only detect that the backup file is missing but NOTHING else

YuryHrytsuk:
https://github.com/kartoza/docker-pg-backup/issues/97

mrnicegyu11:
wontfix, we will move to a backup CI/CD pipeline

PR #440, opened 2023-11-15
Show pg backup logs

YuryHrytsuk:
### Changes
This PR introduces logs for `pg-backup` containers. Before there were no logs in containers logs.

### Discussion about implementation 
* https://github.com/kartoza/docker-pg-backup/issues/97

### It solves issue
* https://github.com/ITISFoundation/osparc-ops-environments/issues/415


services/pg-backup/docker-compose.yml.j2
```diff
@@ -4,6 +4,9 @@ services:
     image: kartoza/pg-backup:14-3.3@sha256:4d1867707b1da879324199fb3c0d542122113bfc50cd2418764efb8f4309d6c1
     volumes:
       - /tank/${MACHINE_FQDN}:/backups
+      - ./custom-backups-cron:/build_data/backups-cron  # https://github.com/kartoza/docker-pg-backup/issues/97#issuecomment-1812504162
```
mrnicegyu11, line 7: You bind-mount volumes here, so pg-backup then has to run on the machine that holds the files. I think with our current CI/CD pipeline no machine even has those files on its local harddrive. So I am not sure if this will work (you can try it out though, maybe it works)



You could use docker configs as well I guess

mrnicegyu11: I also see that the custom-backups script stem from discussion in the issue that you linked. Maybe you want to mkae this link more prominent at a comment at the top of some files as well, since this explains why the docker service is configured exactly like this (and not a more "normal looking" way)

mrnicegyu11 (approved):
Thanks, looks good. Maybe check that the bind mounted files work once this is rolled out ;)

services/pg-backup/docker-compose.yml.j2
YuryHrytsuk: Good catch. Thank you!

YuryHrytsuk:
The effort required to implement it is bigger then the profit it brings. So, I am closing.

If you still want to see logs, get inside the running `pg-backup` container and run `cat /var/log/cron/log`

Issue #419, opened 2023-11-07
Force start maintenance page pipeline stage fails


YuryHrytsuk:
### Actual behaviour
Running `force_start_maintenance_page` in a pipeline (deploying to staging or prod) fails

```bash
failed to create service maintenance-page_maintenance_page_hpcsciencestagingsim4lifecloud: Error response from daemon: rpc error: code = InvalidArgument desc = name must be 63 characters or fewer
```
Example pipeline: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/jobs/3756817

### Expected behviour
Running `force_start_maintenance_page`  deploys maintenance page and succeeds

mrnicegyu11:
fixed at least on staging, closing

PR #462, opened 2023-11-27
Rename maintenance page stack - make it shorter

mrnicegyu11:
This addresses the too long stack name in https://github.com/ITISFoundation/osparc-ops-environments/issues/419


YuryHrytsuk (approved):
Let's use a common PR template :wink:

Issue #466, opened 2023-11-28
Remove `rollback_config` from `docker-compose.yml`


mrnicegyu11:
As laid out by @sanderegg , the current `rollback_config` we set in e.g. https://github.com/ITISFoundation/osparc-ops-environments/blob/72cb2ab4bfd5c745d950fb2375fa85f3e1231c98/services/simcore/docker-compose.yml#L12 is really serving no purpose (except maybe to confuse ourselves), since when we release we atm kick out all users. So we do not require any rollback at the time of releasing since for `staging`/`prod` a human will always be present to validate the successful start of the new docker container.

PR #467, opened 2023-11-28
Remove `rollback_config`, add `failure_action: continue` to update_config

mrnicegyu11:
## What do these changes do?
We change the failure_action from `rollback` to `continue`, since rollback only makes sense during no-downtime updates (that we dont have atm) in any case. A rollback during a release is reprieved as confusing by some.

We now switch to the behavior `continue`, which means that after 5 unsuccessful starts of a docker service's replica, the task associated to the replica will stop trying to start containers, *without reverting and starting the previous container*. The task wills stay in state "failed".

I'll add that there is barely any official documentation available for the `failure_action` and thus I cannot provide good links for details. I did my research by reading the swarmkit sourcecode a bit and testing all `failure_actions` on my local machines.
## Related issue/s
https://github.com/ITISFoundation/osparc-ops-environments/issues/466
## Related PR/s
n/a
## Checklist
- [ ] ~~I tested and it works~~
- [x] I tested the behviour of the `failure_actions` on mock docker stacks on my machine

### DIFF
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -7,11 +7,8 @@ services:
       update_config:
         parallelism: 2
         order: start-first
-        failure_action: rollback
+        failure_action: continue
         delay: 10s
-      rollback_config:
-        parallelism: 0
-        order: stop-first
       restart_policy:
         condition: any
         delay: 5s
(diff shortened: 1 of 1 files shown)


matusdrobuliak66 (approved):
I have a mix feelings here, as I think that we should always do steps which are heading to the long term goal which is no downtimes with continues delivery. This step is going against that, but I get that now it causes more confusion than benefits.

PR #470, opened 2023-12-05
Make S4L admin overview grafana dashboard changes persistant

mrnicegyu11:
## What do these changes do?
- Make S4L admin overview grafana dashboard changes persistent as requested by @mguidon 
- Bonus: Fix pre-commit hook (`SIMCORE_IMAGE_TAG` --> `DOCKER_IMAGE_TAG`)
## Related issue/s
n/a
## Related PR/s
n/a
## Checklist

- [x] I tested and it works


scripts/pre-commit-hooks/stack-yml-generated-is-valid/launch.bash
```diff
@@ -42,7 +42,7 @@ source "$repo_basedir"/scripts/logger.bash
 repo_config=$(cat "$repo_basedir"/.config.location)
 set -o allexport
 # shellcheck disable=SC1090
-export SIMCORE_IMAGE_TAG=master-github-latest
+export DOCKER_IMAGE_TAG=master-github-latest
 # shellcheck source=/dev/null.
```
YuryHrytsuk, line 46: Shouldn't this come from repo.config?

mrnicegyu11: @YuryHrytsuk this actually not, this is usually version pinned to the current staging / prod release and comes directly from the CI/CD pipeline. So for this pre-commit hook we need to give it some value otherwise the `set +nounset` in the bash scripts will error out since a variable is unset ;)

PR #478, opened 2023-12-13
Use custom certs on Dalco & TIP

YuryHrytsuk:
## What do these changes do?
* Make [dalco, tip] traefik work with custom defined certificates
* Drops previously-used `myresolver` cert resolver

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/394

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/225
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/226

## Checklist

- [ ] I tested and it works
  - It is tested on master. For Dalco and TIP I just replicated the changes. We will see during the release
  - I also run `make up-dalco` for some of the services locally and it worked
  - `make .stack.<name>.yml-dalco`  & `make .stack.<name>.yml-public` were run for every service and worked

PR #479, opened 2023-12-13
Fix dalco tip traefik volumes

YuryHrytsuk:
## What do these changes do?
The certificate should be mounted via volumes inside traefik containers. These changes add the necessary volume mount

## Related PR/s
* https://github.com/ITISFoundation/osparc-ops-environments/pull/478

## Checklist

- [ ] I tested and it works --> to be tested during release


mrnicegyu11:
thx!

Issue #512, opened 2024-01-15
graylog in master periodically unhealthy and restarting


pcrespov:
![image](https://github.com/ITISFoundation/osparc-ops-environments/assets/32402063/0de226a4-0d74-4a0f-a2a9-6a10e3a44122)

Seems a configuration problem
![image](https://github.com/ITISFoundation/osparc-ops-environments/assets/32402063/5a74c577-9391-4652-bc43-0118d25b8df0)

pcrespov:
This is happening again in master ... fails, get unhealthy  and the swarm shuts it down

<img width="1399" alt="image" src="https://github.com/ITISFoundation/osparc-ops-environments/assets/32402063/47dcb90c-d785-4c48-84ae-ae587f2d43f8">

PR #518, opened 2024-01-19
Graylog remove slack notifications

YuryHrytsuk:
## What do these changes do?
* Remove the notifications since we don't use them and they are broken

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/512

## Checklist

- [X] I tested and it works


scripts/pre-commit-hooks/stack-yml-generated-is-valid/launch.bash
```diff
@@ -51,17 +51,11 @@ rm "$repo_basedir"/docker-compose.yml 2>/dev/null || true
 rm "$repo_basedir"/.env-devel 2>/dev/null || true
 # via https://stackoverflow.com/a/2466755
 rm -rf osparc-simcore 2>/dev/null || true
-git clone "${GIT_SIMCORE_REPO_URL}"
+git clone https://github.com/ITISFoundation/osparc-simcore
```
pcrespov, line 54: TIP:  When you are interested only in the latest version & branch of a repo, do instead a shallow clone ``git clone --depth=1``. It would be faster and needs to download much less

SEE https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---depthltdepthgt

YuryHrytsuk: Done. Thank you for the remark!

PR #520, opened 2024-01-19
Reduce overprovisioned resource reservations

YuryHrytsuk:
## What do these changes do?
Reduce overprovisioned resource reservations

## Checklist

- [x] I tested and it works


mrnicegyu11:
thanks for this looked fine :--)

PR #522, opened 2024-01-24
Bump storage memory limit

YuryHrytsuk:
## What do these changes do?
Set simcore storage memory limit to `3G` after Melanie killed storage (OOM) by opening a big study

fyi: @matusdrobuliak66 @sanderegg @pcrespov 

## Checklist

- [X] I tested and it works


mrnicegyu11 (approved):

Konohana0608:
Hey guys
Thank you so much for your effort and for raising the download limit~~ I have tried to use it now with the students. Out of 5 people, 3 were able to get it in the first attempt, the remaining two had one error and then it worked in the second attempt. 
Most likely it was because they started downloading at the same time. 
Thanks again for the support

PR #525, opened 2024-01-25
Add missing DNS config and warning in docker-compose.aws.yml

mrnicegyu11:
## What do these changes do?
 Add missing DNS config in docker-compose.aws.yml of appmotion gateway
 Add DNS warning comment
## Checklist

- [x] I rested and it lurks


sanderegg (approved):
this is actually perfectly legal in docker compose world.
but if it fixes the issue then so be it if it is anyway duplicated.

YuryHrytsuk (approved):

PR #526, opened 2024-01-25
Add Dask monitoring

sanderegg:
## What do these changes do?
The dask backend provides a /metrics entrypoint
This might be useful to check what is going on in the computational backend
## Checklist

Actually would be interested in how you test beforehand @YuryHrytsuk @mrnicegyu11 thanks
- [ ] I tested and it works


mrnicegyu11 (approved):
lgtm should work out of the box after `force_redeploying` the monitoring stack (since a `docker config` file is changed here.

re radiobutton: "I tested it and it works". @YuryHrytsuk might be able to say more he introduced it. For me personally, I ~~might have lied sometimes~~ was always very honest here in the past :O After all, when can one be *really* sure that it works? 

In this specific case, you could start the thing locally (for example `make-up` manually traefik, monitoring, simcore), and you may then check in prometheus if the endpoint is correctly discovered and scraped (prometheus has a "service discovery" overview page for this). I can show you how if you want 🍷

YuryHrytsuk (approved):
Go ahead and merge it

PR #528, opened 2024-01-29
Bump dask resource limits

YuryHrytsuk:
## Checklist

- [X] I tested and it works


mrnicegyu11 (approved):

sanderegg (approved):
the same should be done in AWS staging

YuryHrytsuk:
> the same should be done in AWS staging

This is used for all deployments.

PR #565, opened 2024-02-16
Fix prometheus reservations

YuryHrytsuk:
## Checklist

- [X] I tested and it works


mrnicegyu11 (approved):
I think `          memory: 24576M` was there for have prometheus handle queries that go over long times, like many months, without being killed by the OOM killer.

Maybe have a chat with eli and see if her metrics-collection scripts might be impacted by this, not sure

--> fine for me htx

YuryHrytsuk:
> I think ` memory: 24576M` was there for have prometheus handle queries that go over long times, like many months, without being killed by the OOM killer.
> 
> Maybe have a chat with eli and see if her metrics-collection scripts might be impacted by this, not sure
> 
> --> fine for me htx

I brought back the limits :+1:  Good catch

PR #567, opened 2024-02-19
Change deploy policy

GitHK:
## What do these changes do?

In order to avoid issues with existing services, `director-v2` must first be killed and then started.

## Related PR/s

- fixes https://github.com/ITISFoundation/osparc-ops-environments/issues/563
## Checklist

- [ ] I tested and it works

### DIFF
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -396,7 +396,7 @@ services:
     deploy:
       update_config:
         parallelism: 2
-        order: start-first
+        order: stop-first
         failure_action: continue
         delay: 10s
       restart_policy:


YuryHrytsuk (approved):

services/simcore/docker-compose.yml
```diff
@@ -396,7 +396,7 @@ services:
     deploy:
       update_config:
         parallelism: 2
-        order: start-first
+        order: stop-first
```
pcrespov, line 399: i wonder how does this coordinate with the rest of the services that follow `start-first` policy. What happens if one of the latter  (i.e. newly started) communicates with the directorv2 (being stopped)

Q: what does `parallelism:2` mean in this context?

GitHK: - If they try to communicate with the service. Clearly they will get an error. It's up to the other services to deal with it. I will remind that currently `director-v2` is not restartable.
- Restart parallelism is related to the number of copies that are allowed to start at a give time while the service is being scaled. Since we only have one copy this is OK.

pcrespov: - Did you check that the other services "deal with it"?
- is it then necessary?

pcrespov (approved):

services/simcore/docker-compose.yml
GitHK: - The other services might not deal with it. But it's OK since this will happen during maintenance or by operator intervention.
- Yes avoid issues with director-v2 instances possibly doing the same operation twice.

PR #573, opened 2024-02-22
Access storage via traefik

YuryHrytsuk:
## What do these changes do?
This allows an alternative way to connect to storage bypassing some of the docker network machinery. This way to access storage should be more robust

## Related issue/s
* https://github.com/ITISFoundation/osparc-issues/issues/1264

## Checklist

- [X] I tested and it works


sanderegg (approved):

matusdrobuliak66 (approved):
thanks :+1: We should not forget about this one (this should not be a long term solution to this problem, because of security reasons)

PR #589, opened 2024-02-28
Avoid autoscaled nodes

YuryHrytsuk:
## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/585

## Checklist
- [X] I tested and it works


mrnicegyu11 (approved):
top thx

services/admin-panels/docker-compose.aws.yml
```diff
@@ -3,3 +3,7 @@ services:
   adminpanels:
     dns: # Add this always for AWS, otherwise we get "No such image: " for docker services
       8.8.8.8
+    deploy:
+      placement:
+        constraints:
+          - node.role == manager
```
sanderegg, line 9: question why the adminpanels must go on the manager? do they need the docker socket? if yes why is this not defined on the base docker-compose file instead of the AWS one?

sanderegg (commented):
thanks. there is just one question.

services/admin-panels/docker-compose.aws.yml
YuryHrytsuk: This is a good question. `admin-panels` are already set up to go on `manager`. @mrnicegyu11 do you know if there is any necessity in placing `admin-panels` on `manager` nodes?

mrnicegyu11: I dont think there is a reason anymore since we use docker configs. In the past there were some bind-mounted files so likely this is where the manager-constraint originally came from. I think it can be removed

YuryHrytsuk: Moved to ops node

PR #596, opened 2024-03-06
Add endpoint statistics to simcore service dashboard

bisgaard-itis:
## What do these changes do?
Changes are done to the simcore service dashboard.
- Ensure average request/response times/sizes are averaged over services instances to make for numbers which are easier to interpret.
- Introduce a table of statistics (computed since `t=-"infinity"`) for requests duration: One table containing the expected request duration and one containing an estimated 95% quantile. The idea is that these tables can be used to estimate what are meaningful timeout parameters when communicating between the services and also to understand when certain endpoints should be converted to long running taks.
image

## Checklist

- [x] I tested and it works


mrnicegyu11 (approved):
thx

pcrespov (approved):
🎉

PR #607, opened 2024-03-18
606 aiohttp simcore monitoring

### LINKED ISSUE #606: add dashboard for monitoring aiohttp simcore services

bisgaard-itis:
## What do these changes do?
- Add dashboard for monitoring Simcore services based on aiohttp

## Related issue/s
- resolves #606 
## Checklist

- [x] I tested and it works


mrnicegyu11 (approved):
This looks good, so do I understand it correctly that you have built a custom dashboard which visualized prometheus metrics that we any had for `aiohttp`? Thanks!

bisgaard-itis:
> This looks good, so do I understand it correctly that you have built a custom dashboard which visualized prometheus metrics that we any had for `aiohttp`? Thanks!

Exactly, when starting setting this up I realized that the `metrics` endpoint is already exposed for the aiohttp simcore services but the metrics collected are in a slightly different format to the ones we collect for fastapi. So I decided to create a separate dashboard for them as it seemed like the easiest solution.

PR #609, opened 2024-03-18
fix sed regex in docker-compose helper bashscript

mrnicegyu11:
## What do these changes do?
Change regex.
Prior `sed` regex assumed `name:` came on line one. It can come in any line and needs to be removed.
The updated regex will remove all lines that start with `name:  `, although we expect one at most.
This is needed to deploy traefik to dalco.
## Checklist

- [x] I tested and it works


mrnicegyu11:
doesnt apply, read: https://github.com/docker/compose/issues/9306#issuecomment-1689720182

what is mentioned is the root issue, we need to switch to `docker stack config`

PR #610, opened 2024-03-19
Fix: Use `docker stack config`

mrnicegyu11:
## What do these changes do?
We have been using custom hackery to mitigate that `docker compose config` is not compatble with docker swarm.

Luckily, the docker guys have a new command that works: `docker stack config`. This PR adapts our deploy-scripts to use this, in line with https://github.com/docker/compose/issues/9306#issuecomment-1689720182

During the docker engine update on dalco this surfaced, since deploying docker stacks to dalco was not possible anymore with our previous tooling. I really dont know why it continued to work on aws-master, etc. I am still trying to reproduce on that end.

CC @sanderegg how are you currently handling this in simcore?

Bonus:
Fully deprecates docker-compose v1.
## Related issue/s
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/jobs/4689324 failed CD run, reproducible, I traces this to the docker compose issue mentioned above
## Checklist

- [x] I tested and it works


scripts/docker-compose-config.bash
```diff
@@ -57,19 +57,20 @@ if docker compose version --short | grep --quiet "^2\." ; then
...
-
+# shellcheck disable=SC2002
   docker_command="\
-docker \
---log-level=ERROR \
-compose \
---env-file ${env_file}"
+set -o allexport && \
+export $(cat "${env_file}" | sort) && set +o allexport && \
+docker stack config"
```
sanderegg, line 64: :hot:
that is new. what are the implications? and does it use the .env file as well?
do we need to copy that also in osparc-simcore? since this script comes from there. you totally remove docker-compose v1 as well right? that is good.

sanderegg (approved):
few questions. looks great.

scripts/docker-compose-config.bash
mrnicegyu11: this bash-code sources the `.env` file, which is how the docker guys expect you to use "docker stack config", I guess you should also use `docker stack config` in osparc-simcore. I actually copied the bash script from there originally, so you might get lucky and it is possible to just replace the one in `osparc-simcore` with this version

PR #612, opened 2024-03-21
add services into the monitored network

sanderegg:
## What do these changes do?
- adds autoscaling, clusters-keeper, payments, invitations, dynamic-schldr services into the monitored network so they can provide metrics
## Checklist

- [ ] I tested and it works


bisgaard-itis (approved):
very cool, thanks a lot for cleaning up my half-done work

PR #619, opened 2024-03-26
Monitoring: add auto-discovery for computational clusters

sanderegg:
## What do these changes do?
- [x] adapt Makefile to configure prometheus for auto-discovery
- [x] currently scans any manager machine as defined by the clusters-keeper ENV variables
- [x] scans 1 region only at the moment
 
## Checklist

- [x] I tested and it works


services/monitoring/prometheus/prometheus-aws.yml
```diff
@@ -0,0 +1,38 @@
+scrape_configs:
+  - job_name: 'federate-computational-clusters-ec2'
+    honor_labels: true
+    metrics_path: '/federate'
```
mrnicegyu11, line 4: very nice, bu we should have basic auth nevertheless there as well

services/monitoring/prometheus/prometheus-aws.yml
```diff
@@ -0,0 +1,38 @@
...
+    params:
+      # we match all metrics for now
+      'match[]':
```
mrnicegyu11, line 7: I have no clue, are you sure this should be quoted?

services/monitoring/prometheus/prometheus-aws.yml
```diff
@@ -0,0 +1,38 @@
...
+        - '{job=~"^.+$"}'
+    ec2_sd_configs:
+      - region: ${CLUSTERS_KEEPER_EC2_REGION_NAME}
+        access_key: ${CLUSTERS_KEEPER_EC2_ACCESS_KEY_ID}
+        secret_key: ${CLUSTERS_KEEPER_EC2_SECRET_ACCESS_KEY}
+        # profile: "<aws_profile>" # optional, if using instance role or profile
+        port: 9090
```
mrnicegyu11, line 14: if you dont use basic auth, potentially obscure this to use a non-standard port

mrnicegyu11 (changes requested):
Very Nice, just some minors

services/monitoring/prometheus/prometheus-aws.yml
sanderegg: yes, do you have experiment on how to add the basic auth stuff?

services/monitoring/prometheus/prometheus-aws.yml
sanderegg: well it worked on my local machine. now I am not sure yet if it will work from an EC2 machine directly.

services/monitoring/prometheus/prometheus-aws.yml
sanderegg: and the auth I guess will need to be set into osparc-simcore as well

mrnicegyu11 (approved):
let's statr like this in case

services/monitoring/prometheus/prometheus-aws.yml
sanderegg: as discussed it needs 2 PRs:
will do one now in osparc-simcore to setup a password

PR #646, opened 2024-04-09
⚡️ Reduce number of git-repo interconnections (`3 --> 1`)

mrnicegyu11:
## What do these changes do?
@YuryHrytsuk recently nicely refactored the grafana dashboards (which are per-deployment assets) from this repo into the ops-config repo. This is the correct place for them to be. In the process however, the interconnection between the git-repos was made more complicated, since 2 new "linker" files to specify directories containing grafana dashboards were introduced. This makes setting up the interconnected ops-repos harder for backend developers.

This PR refactors this and again reduced the number of linker files to 1, the known and old ".config.location" file that specifies the location of the `repo.config` file.

This PR then, instead of requiring more paths explicitly, will implicitly derive the location of grafana dashboard containing directories in ops-config *relative to the repo.config filepath*. 

feedback welcome!
## Checklist

- [x] I tested and it works

### DIFF
--- a/services/monitoring/Makefile
+++ b/services/monitoring/Makefile
@@ -173,23 +173,16 @@ smokeping_prober_config.yaml: smokeping_prober_config.yaml.j2 ${REPO_CONFIG_LOCA
 	$(call jinja, $<, $@);
 
 .PHONY: grafana/assets
-grafana/assets: .config.dir .shared_config.dir
-	$(eval CONFIG_DIR=$(shell cat $(REPO_BASE_DIR)/.config.dir))
-	$(eval SHARED_CONFIG_DIR=$(shell cat $(REPO_BASE_DIR)/.shared_config.dir))
-
-	$(eval GRAFANA_ASSETS_DIR=$(CONFIG_DIR)/assets/grafana)
-	@if [ ! -d $(GRAFANA_ASSETS_DIR) ]; then \
-		echo "Error: folder does not exist $(GRAFANA_ASSETS_DIR)" >&2; \
+grafana/assets: ${REPO_CONFIG_LOCATION}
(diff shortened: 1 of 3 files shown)


services/monitoring/Makefile
```diff
@@ -173,23 +173,16 @@ smokeping_prober_config.yaml: smokeping_prober_config.yaml.j2 ${REPO_CONFIG_LOCA
...
 .PHONY: grafana/assets
-grafana/assets: .config.dir .shared_config.dir
-	$(eval CONFIG_DIR=$(shell cat $(REPO_BASE_DIR)/.config.dir))
-	$(eval SHARED_CONFIG_DIR=$(shell cat $(REPO_BASE_DIR)/.shared_config.dir))
-
-	$(eval GRAFANA_ASSETS_DIR=$(CONFIG_DIR)/assets/grafana)
-	@if [ ! -d $(GRAFANA_ASSETS_DIR) ]; then \
-		echo "Error: folder does not exist $(GRAFANA_ASSETS_DIR)" >&2; \
+grafana/assets: ${REPO_CONFIG_LOCATION}
+	@if [ ! -d "$(shell dirname ${REPO_CONFIG_LOCATION})/assets/grafana" ]; then \
```
YuryHrytsuk, line 177: repo config location points to a file not directory.

YuryHrytsuk: oh, there is a `dirname` here. 

Why not using `.config.dir` instead of `.config.location`? By doing so we won't need to use `dirname` everytime we need to set up a path

mrnicegyu11: Since dirname can always be used, it is not really all that relevant if we point to a file or directory in my eyes. It would be good to have a pre-commit script that checks all assumptions about the folder-structure, otherwise we will have issues with these deep-nested directory traversals in the future.... Anyway, doing it now would require changes everywhere, so in case we do it I would prefer it in a separate PR.

PR #650, opened 2024-04-10
registry: remove hardcoded aws region

YuryHrytsuk:
## Checklist

- [X] I tested and it works


mrnicegyu11:
puh good find, nasty

PR #661, opened 2024-05-03
adds new web-api service starting with two replicas

pcrespov:
## What do these changes do?

- adds deploy configs for `wb-api-server`, a new service is introduced in https://github.com/ITISFoundation/osparc-simcore/pull/5767. 

- NOTE that this configuration needs to be moved to `docker-compose.yml` as soon as  https://github.com/ITISFoundation/osparc-simcore/pull/5767 reaches staging and production

## Related PR/s

- This needs to get into master **after** PR https://github.com/ITISFoundation/osparc-simcore/pull/5767, right?

## Checklist

- [x] I tested and it works
image

### DIFF
--- a/services/simcore/docker-compose.deploy.master.yml
+++ b/services/simcore/docker-compose.deploy.master.yml
@@ -69,7 +69,11 @@ services:
 
   webserver:
     deploy:
-      replicas: 6
+      replicas: 3
+
+  wb-api-server:
+    deploy:
+      replicas: 3
 
   storage:
     deploy:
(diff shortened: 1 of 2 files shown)


YuryHrytsuk (approved):

mrnicegyu11 (changes requested):
Hi Pedro this looks good but in order to not run into weird issues during releases to stag/prod, I would highly recommend adding your changes to `services/simcore/docker-compose.yml` instead of `services/simcore/docker-compose.deploy.master.yml`. The latter file is usually only used in case we need special (dedicated) resource constraints or volume paths for certain deployments, such as to handle constraint resources on dalco etc. Your code-changes are generic and can - in my view - be propagated anywhere, so `docker-compose.yml` is the most straight-forward place to put these changes :--)

mrnicegyu11 (approved):
thanks, all good!

Issue #665, opened 2024-05-08
monitoring internal network IP exhaustion


YuryHrytsuk:
This happens on S4L every week. The Network size is too small compared to the number of containers we need to run. 

Possible solutions:
* drain nodes before removing then [proper solution] @sanderegg 
* bigger [external] network managed manually
* change docker swarm default subnets size
* get rid of internal monitoring network 
```[tasklist]
### Tasks
- [ ] https://github.com/ITISFoundation/osparc-simcore/issues/5827
- [ ] https://github.com/ITISFoundation/osparc-ops-environments/pull/674
```

sanderegg:
all tasks implemented

PR #674, opened 2024-05-15
Remove internal monitoring network

YuryHrytsuk:
## What do these changes do?
After review with @sanderegg , we figured out that internal network is not really needed and can be removed. Esp. taking into account the problem it causes after being full (IP Exhaustion)

bonus: 
* removed unused `grafana_image_renderer`

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/665

## Checklist

- [x] I tested and it works (partially tested)


mrnicegyu11 (approved):
I love removing sins from the past, thanks a lot! :fire:

Issue #680, opened 2024-05-27
Bug - `Make` broken on some linux flavors 


mrnicegyu11:
See:  https://github.com/ITISFoundation/osparc-ops-environments/pull/681

PR #681, opened 2024-05-27
Fix makefile issues

mrnicegyu11:
### Makefile variable issues
Both @sanderegg and me were unable to run a simple `make` command in this repository. The issue was fixed for me by setting make-variables with `:=` syntax instead of `=`. 

#### `:=` is the "Simple or Immediate Assignment" 
This operator assigns a value to a variable and immediately expands it. The value is fixed at the time of assignment and does not change even if other variables it depends on change later.

#### `=` is the Recursive Assignment:
Behavior: This operator assigns a value to a variable, but the value is not immediately expanded. Instead, the value is expanded each time the variable is used.

(Explanations in this text by LLMs, no factchecking, hallucination warning,)

### `hostname` issues 
In the `common.Makefile` a helper functinon to determine the private-ip of the machine that the command runs on is used. This function calls the linux binary `hostname` in a  shell-call. It turns out that there are multiple implementations of `hostname` that have *different versioning schemes` and are `incompatabile with each other`.

The `hostname` command originally is part of the old `net-tools` package, which is pretty much considered outdated and unmaintained these days. Most modern GNU/Linux distributions have moved on to a newer version provided by GNU/inetutils (see: https://archlinux.org/news/hostname-utility-moved-from-net-tools-to-inetutils/, https://forum.archlinux.de/d/19944-command-not-found-hostname/3 ). Now, here's where it gets a bit tricky. Over in the Debian world, a contributor named Michael Meskes has created a special version of the `hostname` command just for Debian (see https://net-tools.sourceforge.io/ , https://metadata.ftp-master.debian.org/changelogs//main/h/hostname/hostname_3.23+nmu1_changelog ). This version has its own versioning scheme and isn't compatible with the "official" GNU/inetutils version. One big difference is that the `-I` option (which shows all IP addresses of the host) isn't supported by the GNU/inetutils or the original `net-tools` version of `hostname`. So, if you rely on `hostname -I`, you might run into compatibility issues or unexpected behavior, especially if your script ends up running on different systems. To keep things smooth and avoid any headaches, I propose to instead use the `-i` argument, that works well for me.

## Related issue/s
https://github.com/ITISFoundation/osparc-ops-environments/issues/680
## Checklist

- [x] I tested and it works

### DIFF
--- a/scripts/common.Makefile
+++ b/scripts/common.Makefile
@@ -34,12 +34,12 @@ endif
 
 
 # Parse the different FQDNS in the repo.config (REPO_CONFIG_LOCATION) and convert them into traefik typo and add the service rule
-export DEPLOYMENT_FQDNS_CAPTURE_TRAEFIK_RULE_CATCHALL=$(shell set -o allexport; \
+export DEPLOYMENT_FQDNS_CAPTURE_TRAEFIK_RULE_CATCHALL:=$(shell set -o allexport; \
 	source $(REPO_CONFIG_LOCATION); \
 	if [ -z "$${DEPLOYMENT_FQDNS}" ]; then \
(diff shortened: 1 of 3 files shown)


sanderegg (approved):

YuryHrytsuk:
`hostname -i` and `hostname -I` return different results in my shell.

```bash
-i, --ip-address
-I, --all-ip-addresses
```

Does it make any difference for our implementation considering the switch?

YuryHrytsuk (approved):

mrnicegyu11:
Good point thanks, I tested it and it works

PR #683, opened 2024-05-30
[AWS]: use EC2 private IP instead of private DNS name as it does not resolve inside the containers

sanderegg:
## Checklist

- [ ] I tested and it works


YuryHrytsuk (approved):
Thank you!

PR #686, opened 2024-06-03
mount efs docker volume to efs-guardian

matusdrobuliak66:
## What do these changes do?
- mount efs docker volume to efs-guardian
## Related issue/s
- relates https://github.com/ITISFoundation/osparc-issues/issues/1313
- relates https://github.com/ITISFoundation/osparc-simcore/issues/5872

## Checklist

- [ ] I tested and it works

### DIFF
--- a/services/simcore/docker-compose.deploy.aws.yml
+++ b/services/simcore/docker-compose.deploy.aws.yml
@@ -7,6 +7,10 @@ services:
       placement:
         constraints:
           - node.role == worker
+  efs-guardian:
+    volumes:
+      - efs_volume:/data/efs
+
   clusters-keeper:
     deploy:
       replicas: 1
@@ -58,3 +62,10 @@ services:
   payments:
     deploy:
       replicas: 1
+
+volumes:
+   efs_volume:
+       driver_opts:
+           type: nfs
+           o: addr=${EFS_DNS_NAME},rw,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport
+           device: :/
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -532,6 +532,34 @@ services:
           memory: 128M
           cpus: '0.1'
 
+  efs-guardian:
+    networks:
+      - monitored
+      - public
+    hostname: "{{.Service.Name}}"
+    deploy:
+      replicas: ${SIMCORE_EFS_GUARDIAN_REPLICAS}
+      update_config:
+        parallelism: 2
+        order: start-first
+        failure_action: continue
+        delay: 10s
+      restart_policy:
+        condition: any
+        delay: 5s
+        max_attempts: 3
+        window: 120s
+      placement:
+        constraints:
+          - node.labels.simcore==true
+      resources:
+        limits:
+          memory: 256M
+          cpus: '0.5'
+        reservations:
+          memory: 64M
+          cpus: '0.1'
+
   migration:
     deploy:
       resources:
--- a/services/simcore/docker-compose.yml
+++ b/services/simcore/docker-compose.yml
@@ -533,9 +533,6 @@ services:
           cpus: '0.1'
 
   efs-guardian:
-    networks:
-      - monitored
-      - public
     hostname: "{{.Service.Name}}"
     deploy:
       replicas: ${SIMCORE_EFS_GUARDIAN_REPLICAS}


services/simcore/docker-compose.yml
```diff
@@ -532,6 +532,34 @@ services:
           memory: 128M
           cpus: '0.1'
 
+  efs-guardian:
+    networks:
+      - monitored
+      - public
```
sanderegg, line 538: Question: why are these networks needed?

sanderegg (commented):
just wondering about the networks you put that guy in?
why monitored? does it need to be monitored by prometheus?
why public? is it accessible via traefik?

services/simcore/docker-compose.yml
matusdrobuliak66: No they are not needed, thanks for pointing out. I will remove them.

sanderegg (approved):
thanks

services/simcore/docker-compose.yml
```diff
@@ -532,6 +532,34 @@ services:
...
+        delay: 5s
+        max_attempts: 3
+        window: 120s
+      placement:
+        constraints:
+          - node.labels.simcore==true
+      resources:
+        limits:
+          memory: 256M
+          cpus: '0.5'
```
mrnicegyu11, line 555: just fyi: this might lead to your service being throttled. There is a prometheus metric to detect throttling, let me know if you care

services/simcore/docker-compose.yml
```diff
@@ -532,6 +532,34 @@ services:
           memory: 128M
           cpus: '0.1'
 
+  efs-guardian:
+    networks:
```
mrnicegyu11: Are you sure this service needs to be in the `public` network?

services/simcore/docker-compose.deploy.aws.yml
```diff
@@ -58,3 +62,10 @@ services:
   payments:
     deploy:
       replicas: 1
+
+volumes:
+   efs_volume:
+       driver_opts:
+           type: nfs
+           o: addr=${EFS_DNS_NAME},rw,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport
```
mrnicegyu11, line 70: looks legit

mrnicegyu11 (approved):
feuer frei!

services/simcore/docker-compose.yml
mrnicegyu11: Actually I am also sceptical now about the `monitoring` network, as SAN mentioned as well

matusdrobuliak66: networks were removed :+1:

services/simcore/docker-compose.yml
```diff
@@ -532,6 +532,34 @@ services:
...
+        condition: any
+        delay: 5s
+        max_attempts: 3
+        window: 120s
+      placement:
+        constraints:
+          - node.labels.simcore==true
+      resources:
+        limits:
+          memory: 256M
```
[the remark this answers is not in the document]
YuryHrytsuk: We measure this metrics on a running service. Copying it is OK to start with and then we can adjust.

services/simcore/docker-compose.yml
```diff
@@ -532,6 +532,31 @@ services:
           memory: 128M
           cpus: '0.1'
 
+  efs-guardian:
+    hostname: "{{.Service.Name}}"
+    deploy:
+      replicas: ${SIMCORE_EFS_GUARDIAN_REPLICAS}
```
YuryHrytsuk, line 538: @mrnicegyu11 I would not use ENV for replicas but just write then in `docker-compose.<deployment>.yaml`. Do you have an opinion here?

services/simcore/docker-compose.deploy.aws.yml
```diff
@@ -7,6 +7,10 @@ services:
       placement:
         constraints:
           - node.role == worker
+  efs-guardian:
+    volumes:
+      - efs_volume:/data/efs
```
YuryHrytsuk, line 12: if you use volumes that are kept on the machine, then you service should be bound strictly to this particular machine. If this is the case, then `simcore` label does not fit since this label is attached to multiple machines. We can discuss ...

YuryHrytsuk (approved):
Thanks! Please consider the comments

services/simcore/docker-compose.yml
matusdrobuliak66: @YuryHrytsuk The issue here is that I cannot differentiate between AWS deployments since you have only one file for that. Therefore I need to differentiate via ENV vars.

services/simcore/docker-compose.deploy.aws.yml
matusdrobuliak66: Ok, I probably copy pasted it, should I remove it then?

matusdrobuliak66: As discussed as this volume is using distributed file system, it can be reused across multiple instances.

Issue #716, opened 2024-07-12
Dockerhub rate limits issue when using Private EC2s


sanderegg:

YuryHrytsuk:
### Literature
* [Docker Hub rate limit](https://docs.docker.com/docker-hub/download-rate-limit/)

YuryHrytsuk:
PR fixing docker login on cluster machines: https://git.speag.com/oSparc/osparc-infra/-/merge_requests/237

YuryHrytsuk:
`Docker log in per request` vs `Private registry as pull through cache to Dockerhub`

**Docker log in per request**
\- need to be done per request (every time)
\- need to log in on every machine 

**Private registry as pull through cache to Dockerhub**
\+ once set up works automatically out-of-the-box with no overhead
\- OPS need to guarantee private registry uptime
\- OPS need to properly loadbalance and scale private registry
\- OPS need to maintain disk space of private registries
\- once registry is down no new service on autoscaled machine can start (based on the assumption image pulls happen on new machines)
\- daemon.json needs to be configured on every machine

Based on this comparison, I decide to go on with `Docker log in per request` approach. Even though I would rather have something working automatically out-of-the-box but the cost is too high

@mrnicegyu11, @sanderegg, let me know what are your thoughts on this matter

YuryHrytsuk:
### Docker swarm and authenticated image pulling. Questions

Does it matter which user docker container process has? (unrelated to docker swarm)
* It does not. As long as you issue docker command with authenticated user and `--with-registry-auth` option, the pulling of images is authneticated

Do I need to issue `docker login` on every docker swarm node?
*  According to [this SO Answer](https://stackoverflow.com/q/53778764/12124525), you do not need to be logged in on every machine. You need to issue docker login on the machine from where you run docker command

Do I need to issue docker login every time I ssh into machine?
* To be answered

Do I need to issue docker login on the machine or docker swarm already have login  information (in case it was run before with --with-registry-auth)?
* To be answered

sanderegg:
> `Docker log in per request` vs `Private registry as pull through cache to Dockerhub`
> 
> **Docker log in per request** - need to be done per request (every time) - need to log in on every machine
> 
[...]

PR #725, opened 2024-07-31
Introduce pull through cache registry

YuryHrytsuk:
## What do these changes do?
* add pull through registry cache within `registry` stack
* use `published port` to be able to reach pull through cache registry

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/716

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/750
* https://git.speag.com/oSparc/osparc-infra/-/merge_requests/240

## Checklist
- [ ] I tested and it works


services/registry/docker-compose.yml
```diff
@@ -66,7 +66,51 @@ services:
...
+    environment:
+      REGISTRY_HTTP_SECRET: ${REGISTRY_PULL_THROUGH_CACHE_HTTP_SECRET}
+      # S3
+      REGISTRY_STORAGE: s3
+      REGISTRY_STORAGE_S3_ACCESSKEY: ${REGISTRY_S3_ACCESS_KEY}
+      REGISTRY_STORAGE_S3_SECRETKEY: ${REGISTRY_S3_SECRET_KEY}
+      REGISTRY_STORAGE_S3_REGIONENDPOINT: "${REGISTRY_S3_ENDPOINT}"
+      REGISTRY_STORAGE_S3_SECURE: "true"
+      REGISTRY_STORAGE_S3_BUCKET: ${REGISTRY_PULL_THROUGH_CACHE_S3_BUCKET}
+      REGISTRY_STORAGE_S3_REGION: ${REGISTRY_AWS_REGION}
```
sanderegg, line 81: Question: what are the implications of this?
the pull-through registry also points to the private registry data correct?
since inside the deploys we only pull (never push), could it be that we could just use that one inside the deployments?
We only need one to push actually when we add images.

services/registry/docker-compose.yml
```diff
@@ -66,7 +66,51 @@ services:
...
+      REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: inmemory
+      REGISTRY_STORAGE_REDIRECT_DISABLE: "true"
```
sanderegg, line 83: what does this do?

services/registry/docker-compose.yml
```diff
@@ -66,7 +66,51 @@ services:
...
+      REGISTRY_HTTP_SECRET: ${REGISTRY_PULL_THROUGH_CACHE_HTTP_SECRET}
+      # S3
+      REGISTRY_STORAGE: s3
+      REGISTRY_STORAGE_S3_ACCESSKEY: ${REGISTRY_S3_ACCESS_KEY}
+      REGISTRY_STORAGE_S3_SECRETKEY: ${REGISTRY_S3_SECRET_KEY}
+      REGISTRY_STORAGE_S3_REGIONENDPOINT: "${REGISTRY_S3_ENDPOINT}"
+      REGISTRY_STORAGE_S3_SECURE: "true"
+      REGISTRY_STORAGE_S3_BUCKET: ${REGISTRY_PULL_THROUGH_CACHE_S3_BUCKET}
+      REGISTRY_STORAGE_S3_REGION: ${REGISTRY_AWS_REGION}
+      REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: inmemory
```
sanderegg, line 82: this is for the pull cache right? are there other options in the end?
if we scale the registry then there will be multiple caches right? thus increasing the number of pulls?

sanderegg: also this can potentially become large? is the underlying EC2 able to sustain the space?

sanderegg (approved):
cool! a few comments and questions

mrnicegyu11:
I think I need a face2face talk for this to reasonably comment. The yaml etc. looks fine :D

services/registry/docker-compose.yml
YuryHrytsuk: Removed this ENV altogether

services/registry/docker-compose.yml
YuryHrytsuk: pull-through registry does not point to private registry data

YuryHrytsuk: I do not fully grasp the "one-side" way. Could you elaborate if you think it is still the case?

PR #726, opened 2024-08-05
Registry ptc use port host mode

YuryHrytsuk:
## What do these changes do?
1. Use host mode port for registry pull through cache to be able to keep it private even if some docker swarm nodes are public
2. Use global mode and run all replicas on OPS node
3. [Done in other repos] Setup Private DNS Records that resolves to private IPs of all OPS nodes

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
yep thx

sanderegg (approved):
crazy! thanks

PR #735, opened 2024-08-13
Remove k8s base64 secret envs

YuryHrytsuk:
## What do these changes do?
We can use `b64enc` filter in `.gotmpl` templates instead. It allows us to reduce number of ENVs for kubernetes

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/772

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
lovely! thx

Issue #730, opened 2024-08-12
master: traefik missing port for registry pull through cache


pcrespov:
- open graylog dashboard in master
- add filter `log_service:traefik_traefik`
- `traefik` service periodically fails in
time="2024-08-12T11:30:43Z" level=error msg="service \"registry-registry-pull-through-cache\" error: port is missing" providerName=docker container=registry-registry-pull-through-cache-****

![image](https://github.com/user-attachments/assets/a6c6cbd1-2385-45d5-a38a-1a3c54a3502c)

Seems to be a misconfiguration

mrnicegyu11:
This is a new feature by @YuryHrytsuk , please comment briefly

YuryHrytsuk:
> This is a new feature by @YuryHrytsuk , please comment briefly

Yes. We do not use `registry-pull-through-cache` in in-house master. I will take care of it soon.

Thanks

PR #740, opened 2024-08-14
Remove traefik labels from registry pull through cache

YuryHrytsuk:
## What do these changes do?
Remove traefik labels from registry-pull-through cache containers as we do not need traefik networking for this service

fyi: @pcrespov 
## Related issue/s
* closes https://github.com/ITISFoundation/osparc-ops-environments/issues/730

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
as discussed, thx

PR #741, opened 2024-08-14
Add otel-collector

mrnicegyu11:
## Checklist

- [ ] I tested and it works


services/jaeger/docker-compose.yml
```diff
@@ -29,7 +28,22 @@ services:
         reservations:
           memory: 64M
           cpus: "0.1"
-
+  otel-collector:
+    image: otel/opentelemetry-collector-contrib:0.105.0
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otel/config.yaml
```
YuryHrytsuk, line 34: Why not docker config?

YuryHrytsuk (approved):

services/jaeger/docker-compose.yml
mrnicegyu11: good point I shall add that

YuryHrytsuk:
A follow up question:
* can we get rid of `JAEGER` docker node label constraint?

This question stems from our desire to be highly available (aka easily restartable and particular node independent).

mrnicegyu11:
> A follow up question:
> 
>     * can we get rid of `JAEGER` docker node label constraint?
> 
> 
> This question stems from our desire to be highly available (aka easily restartable and particular node independent).

Now that the file is mounted into the container as a docker config, jaeger can be considered "stateless" and can float around.

PR #746, opened 2024-08-15
skipping validation check for efs guardian

matusdrobuliak66:
## What do these changes do?
 - skipping validation check for efs guardian, as it cannot start because mounting the AWS Distributed Elastic File System to the runner would be required.
## Checklist

- [ ] I tested and it works


mrnicegyu11 (approved):
thanks a lot!

PR #786, opened 2024-09-03
OPS Traefik: disable traefik v2 core syntax

YuryHrytsuk:
## What do these changes do?
Disable core v2 syntax. This fixed regex in simcore traefik but was forgotten in ops traefik. Disabling v2 core syntax should fix www redirection.

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
thanks a lot!

Issue #241, opened 2023-06-26
Change Redis on AWS to AWS Hosted Redis


mrnicegyu11:
Do this with terraform

mrnicegyu11:
This is a follow up from https://github.com/ITISFoundation/osparc-ops-environments/issues/505#issuecomment-1924029021

mrnicegyu11:
This is blocked and paused in agreement with the team until the sim4life.io release

mrnicegyu11:
This is blocked until https://github.com/ITISFoundation/osparc-simcore/pull/5708

YuryHrytsuk:
### Literature 

* https://aws.amazon.com/blogs/database/configuring-amazon-elasticache-for-redis-for-higher-availability/

PR #810, opened 2024-09-18
AWS Deployments: close redis external port

YuryHrytsuk:
## What do these changes do?
Do not expose self-hosted Redis. There is no need since we only need it for AWS Master which now uses AWS-hosted Redis

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/241

## Checklist

- [ ] I tested and it works

### DIFF
--- a/services/simcore/docker-compose.deploy.aws.yml
+++ b/services/simcore/docker-compose.deploy.aws.yml
@@ -65,19 +65,6 @@ services:
     deploy:
       replicas: 3
 
-  redis:
-    deploy:
-      labels:
-        - traefik.enable=true
-        - traefik.docker.network=${PUBLIC_NETWORK}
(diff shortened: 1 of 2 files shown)


mrnicegyu11 (approved):
Is the AWS memoryDB already in aws-master active? Nice!

YuryHrytsuk:
> Is the AWS memoryDB already in aws-master active? Nice!

And used. We use Redis OSS (not MemoryDB) because it supports multiple databases while MemoryDB does not support multiple Redis Databases.

The reason is MemoryDB is implemented as Redis Cluster and Redis Cluster does not support `SELECT` statement (required for multiple databases) by definition of it.

it is documented here: https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/blob/configs/deployments/osparc-master-zmt.click/terraform/ElasticCache.tf?ref_type=heads#L44-46

PR #823, opened 2024-09-25
jaeger: use ops node label constraint

YuryHrytsuk:
## What do these changes do?

Avoid using specific `jaeger` node label since there is no need to bound jaeger to a node. `ops` node labels will suffice and will make service more restartable by using more generic `ops` node label

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/747

## Checklist
- [X] I tested and it works


mrnicegyu11 (approved):
thanks a lot, looks good!

PR #841, opened 2024-10-17
🔨: Add hardcoded node label addition on local-k8s/kind

mrnicegyu11:
## What do these changes do?
Make sure that adminer starts on osparc.local, for now by manually adding a label-add step.
Very hardcoded, this is a true FIXME. I will open an issue in osparc-ops-environments. I dont know how to do this more automatic in a smart way for now. Do we have a list of valid node labels somewhere that we could iterate and always add to the single-node local-k8s kind cluster? :thinking: 
## Related issue/s
- https://github.com/ITISFoundation/osparc-ops-environments/issues/842
## Checklist

- [x] I tested and it works


YuryHrytsuk:
We can use the nativa Kind configuration for this. I propose to go this way. I can take it over, if you want

https://kind.sigs.k8s.io/docs/user/configuration/#extra-labels

mrnicegyu11:
Yes very good, I will close this MR

Issue #819, opened 2024-09-23
k8s - Add Service Overview (Portainer)


mrnicegyu11:

mrnicegyu11:
done

PR #848, opened 2024-10-21
Add k8s portainer

mrnicegyu11:
## What do these changes do?
Adds k8s portainer to k8s.

~~This will require a sister-PR in osparc-config/e2e-ops for an e2e test before it is merged.~~ is linked below

Bonus:
- Adds basicath for adminer. Might actually not be desirable, I can remove it as well.

Potential improvements are:
- Can the portainer login/password (which is currently manually created during first login) be managed in helmfile gotmpl files, taking the env-vars `PORTAINER_USER` and `PORTAINER_PASSWORD` from `repo.config.template`? This seems not immedeatly possible, see discussions in https://github.com/portainer/k8s/pull/72 and https://github.com/portainer/k8s/pull/71
- We duplicate the strip regex traefik middleware. Can this be avoided?
## Related issue/s
https://github.com/ITISFoundation/osparc-ops-environments/issues/819
## Related PR/s
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1017
## Checklist

- [x] I tested and it works

### DIFF
--- a/charts/adminer/values.yaml.gotmpl
+++ b/charts/adminer/values.yaml.gotmpl
@@ -52,6 +52,7 @@ ingress:
     namespace: {{ .Release.Namespace }}
     cert-manager.io/cluster-issuer: "cert-issuer"
     traefik.ingress.kubernetes.io/router.entrypoints: websecure
+    traefik.ingress.kubernetes.io/router.middlewares: traefik-traefik-basic-auth@kubernetescrd  # namespace + middleware name
   tls:
     - hosts:
         - {{ requiredEnv "K8S_MONITORING_FQDN" }}
--- a/charts/portainer/values.yaml.gotmpl
+++ b/charts/portainer/values.yaml.gotmpl
@@ -1,0 +1,68 @@
+# Default values for adminer.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: portainer/portainer-ce
+  pullPolicy: IfNotPresent
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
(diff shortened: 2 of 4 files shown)


charts/portainer/values.yaml.gotmpl
```diff
@@ -0,0 +1,68 @@
...
+  port: 9000
+
+ingress:
+  enabled: true
+  className: ""
+  annotations:
+    namespace: {{ .Release.Namespace }}
+    cert-manager.io/cluster-issuer: "cert-issuer"
+    traefik.ingress.kubernetes.io/router.entrypoints: websecure
+    traefik.ingress.kubernetes.io/router.middlewares: traefik-traefik-basic-auth@kubernetescrd,traefik-portainer-strip-prefix@kubernetescrd  # namespace + middleware name
```
YuryHrytsuk, line 42: traefik namespace is hard coded. is there a way to use ENV / VAR?

charts/adminer/values.yaml.gotmpl
```diff
@@ -52,6 +52,7 @@ ingress:
     namespace: {{ .Release.Namespace }}
     cert-manager.io/cluster-issuer: "cert-issuer"
     traefik.ingress.kubernetes.io/router.entrypoints: websecure
+    traefik.ingress.kubernetes.io/router.middlewares: traefik-traefik-basic-auth@kubernetescrd  # namespace + middleware name
```
YuryHrytsuk, line 55: Is there a way to reuse traefik namespace via ENV / Variable?

charts/portainer/values.yaml.gotmpl
mrnicegyu11: I think for traefik it is fine that it is hardcoded. Afterall, we specifically want to fetch these middlewares in other namespaces (example: portainer, adminer) from preciceily the traefik namespace. An alias or env would only be helpful if we switch to a different reverse proxy, which I dont forsee...

Thinking about it again, would you find it acceptable now? :--) Which benefit do you see here with an env-var-alias?

PR #854, opened 2024-10-24
✨ Add: k8s S3 CSI to be used with portainer

mrnicegyu11:
## What do these changes do?
- Add S3 CSI to k8s via helm(file)
- Fixes the bug that portainer wont start on both masters
## Related PR/s
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1025
## Checklist
- [x] I tested and it works


charts/portainer/values.yaml.gotmpl
```diff
@@ -18,6 +18,12 @@ serviceAccount:
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: portainer-sa-clusteradmin
+persistence:
+  enabled: true
+  size: "10Gi"
```
YuryHrytsuk, line 23: if this is a max size of the portainer's volume, I think `1Gi` is more then enough. It is 59MB on inhouse-master and 3.8MB on NIH STAG at the moment

YuryHrytsuk (approved):

charts/portainer/values.yaml.gotmpl
mrnicegyu11: true will change thx

PR #858, opened 2024-10-29
Remove `stack-yaml-generated-is-valid` pre-commit hook

YuryHrytsuk:
## What do these changes do?
Remove `stack-yaml-generated-is-valid` pre-commit hook

**Rationale**
This hook relies on `.config.location` and on manually generated `repo.config` in `osparc-config` repository. This pre-commit hook will fail for any new user trying to commit to the `osparc-ops-environments` since these users are unaware of the dependency on `repo.config` from `osparc-config` repository. Moreover, any non-ITIS user cannot even get this repository (hence commit at all) since it is not publicly available.

UPD:
The idea of pre-commit hook relying on physical generated files from other repositories seems to be wrong. This is why in our GitHub CI we cannot run `pre-commit` as is (we have to skip `stack-yaml-generated-is-valid` hook)

## Checklist
- [X] I tested and it works


mrnicegyu11 (approved):
I can agree with what you wrote, this test was useful some years ago when it still worked properly, but not anymore. Thanks!

PR #872, opened 2024-11-13
Sister PR for https://github.com/ITISFoundation/osparc-simcore/pull/6691

mrnicegyu11:
## What do these changes do?
Add interactive services network to opentelemetry provider in jaeger stack.
## Related PR/s
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1079
https://github.com/ITISFoundation/osparc-simcore/pull/6691
## Checklist
- [ ] I tested and it works


YuryHrytsuk:
The problem of this PR is that OTEL now will not run without `interactive_services_subnet` (read `simcore`). 

The advantage if this approach is security (instead of having all services in common network, we put OTEL to each network)

Issue #883, opened 2024-11-20
Python3.11 is not installed on machines


YuryHrytsuk:
```[tasklist]
### Tasks
- [x] Master
- [x] Stag
- [ ] Prod
```

### Related PRs
* https://git.speag.com/oSparc/osparc-infra/-/merge_requests/262

PR #894, opened 2024-11-27
🐛 Fix: Make `osparc.local` functional again - various minor fixes

mrnicegyu11:
## Checklist
- [ ] I tested and it works


.python-version
```diff
@@ -1 +1 @@
-3.11
+3.10
```
YuryHrytsuk, line 1: we use python 3.11 in `simcore`, `osparc-config` and `osparc-infra` repositores. We also use it on our server machines (check [this Issue](https://github.com/ITISFoundation/osparc-ops-environments/issues/883)).

Is there a specific reason to use 3.10 in `osparc-ops`?

scripts/create-s3-bucket.bash
```diff
@@ -11,5 +11,5 @@ IFS=$'\n\t'
 docker run \
 -v /etc/ssl/certs:/etc/ssl/certs:ro \
 --network host \
---env MC_HOST_local="https://${S3_ACCESS_KEY}:${S3_SECRET_KEY}@${S3_ENDPOINT}" \
-minio/mc:RELEASE.2023-06-19T19-31-19Z mb --ignore-existing local/"$1"
+--env MC_HOST_local="https://${S3_ACCESS_KEY}:${S3_SECRET_KEY}@${STORAGE_DOMAIN}" \
```
YuryHrytsuk, line 14: Why change to `STORAGE_DOMAIN`? The script is called `create-s3-bucket.bash` and `S3_ENDPOINT` seems to be a more appropriate candidate in this context

scripts/deployments/deploy_everything_locally.bash
```diff
@@ -235,6 +235,9 @@ if [ "$start_opsstack" -eq 0 ]; then
     # -------------------------------- ADMIN-PANELS -------------------------------
 
     log_info "starting admin-panels..."
+    # Check if the stack 'admin-panels' exists and delete it if it does
+    # shellcheck disable=2015
+    docker stack ls | grep -q admin-panels && docker stack rm admin-panels >/dev/null 2>&1 || true
```
YuryHrytsuk, line 240: Why do we need to delete `admin-panels` specifically before starting it while all other services in this script are simply started?

Is there some specific corner case?

services/traefik/docker-compose.local.yml
```diff
@@ -11,6 +11,13 @@ services:
       - "--api.dashboard=true"
       - "--log.level=${OPS_TRAEFIK_LOGLEVEL}"
       - "--accesslog=true"
+      - "--accesslog.format=json"
+      - "--accesslog.fields.defaultmode=keep"
+      - "--accesslog.fields.names.ClientUsername=keep"
+      - "--accesslog.fields.headers.defaultmode=keep"
+      - "--accesslog.fields.headers.names.User-Agent=keep"
+      - "--accesslog.fields.headers.names.Authorization=drop"
+      - "--accesslog.fields.headers.names.Content-Type=keep"
```
YuryHrytsuk, line 20: Is it for extensive debug purposes pointed out in PR's description?

.pylintrc
```diff
@@ -58,7 +58,7 @@ ignore-paths=^.*\\generated_models\\.*$|^.*/generated_models/.*$
 # Files or directories matching the regex patterns are skipped. The regex
 # matches against base names, not paths. The default value ignores Emacs file
 # locks
-ignore-patterns=venv,.venv
+ignore-patterns=venv,.venv,jupyter_server_config.py
```
YuryHrytsuk, line 61: Question: is pylint a part of pre-commit? pre-commit has been green :thinking:

YuryHrytsuk (approved):
Thank you!

Left a few comments

.python-version
mrnicegyu11: Good info thanks, I think something didnt work with 3.11 and I simply reverted to 3.10. But in this case it is a good moment to fix the issues and move to 3.11. I will have another look ;)

scripts/create-s3-bucket.bash
mrnicegyu11: `STORAGE_DOMAIN` and `S3_ENDPOINT` encode the same domain, but afaik `S3_ENDPOINT` also contains the prefix `https://`. This is just a quick change of env-vars that are anyway available to make the script work. The script didnt work anymore with `S3_ENDPOINT`

scripts/deployments/deploy_everything_locally.bash
mrnicegyu11: Yes, somehow the jinja2-templating logic that automatically adds jupyter-notebooks that it finds in a folder as `docker configs`. Docker always detects the docker config objects as changed and complains that he cant update the service if docker configs change. This adresses it in a bit shitty manner (always cleanly recreating), but it should be fine for the `osparc.local` usecase I think.

It fixes an issue and adds robustness, at the cost of being a bit slower by first deleting and then recreating, but only on the local deploy.

services/traefik/docker-compose.local.yml
mrnicegyu11: Exactly. I can also leave it away, but the logs are useful and usually one starts `osparc.local` to only debug or develop

.pylintrc
mrnicegyu11: Yes, the pre-commit hook always failed for me due to this file, but it only runs of `.py` files are modified so you might not have observed the failure in a while? not sure.

.python-version
mrnicegyu11: 3.11 works here just fine, going back to 3.11 :--) thanks!

.pylintrc
YuryHrytsuk: In CI we run `pre-commit run --all-files` which I thought means running all `pre-commit` against all files.

scripts/create-s3-bucket.bash
YuryHrytsuk: This is true

scripts/deployments/deploy_everything_locally.bash
YuryHrytsuk: But why does not it happen in running deployments?

.pylintrc
mrnicegyu11: I think this depends on the specific pre-commit hook, but yes I would argue gitignored files should usually not be checked by pre-commit. In this case even after digging I am not sure what is going on, but this anyway fixes it and doesnt hurt ;)

scripts/deployments/deploy_everything_locally.bash
mrnicegyu11: I think it is related to the `jupyter_server_config.py` on my machine somehow. I have not seen it in deployments, only on my local machine. Might be again some custom ubuntu tooling, where I have the GNU version, or something along those lines?

Issue #855, opened 2024-10-25
Kubernets: introduce production ready volume management


YuryHrytsuk:

YuryHrytsuk:
## Literature
* [Container Storage Interface (CSI)](https://github.com/container-storage-interface/spec/blob/master/spec.md)

YuryHrytsuk:
## Multi-Replicated Statefulset and PVs
* https://stackoverflow.com/questions/72250853/what-happens-when-we-create-stateful-set-with-many-replicas-with-one-pvc-in-kube

PR #916, opened 2024-12-30
AWS Kubernetes: prod volume management

YuryHrytsuk:
## What do these changes do?
* Add [aws-ebs-csi](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) support
* Add support multiple storage classes (s3 or ebs) via values.yaml

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/855

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1176

## Checklist
- [X] I tested and it works

### DIFF
--- a/charts/aws-ebs-csi-driver/values.yaml.gotmpl
+++ b/charts/aws-ebs-csi-driver/values.yaml.gotmpl
@@ -1,0 +1,21 @@
+# Relies on ec2 instance profile attached to K8s EC2.
+# It lets ebs csi driver to manage EBS volumes
+
+image:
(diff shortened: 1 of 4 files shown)


mrnicegyu11:
Very nice, thanks a lot. Minor naming suggestion: `pv` I guess means "pivate-volume"? I am not sure this is intuituive... I guess volume would be enough, or the fullname not the abbreviation `pv` ;) But it is fine for the time being :--)

PR #929, opened 2025-01-09
Fix pre-commit step in CI 

YuryHrytsuk:
## What do these changes do?
* There was a [bug introduced](https://github.com/hadialqattan/pycln/issues/249) in hook. The fix comes in a later version
* Use pre-commit github native action

## Checklist
- [X] I tested and it works


mrnicegyu11 (approved):
thanks, perfect

Issue #949, opened 2025-01-27
Add automated sustainable workflow to update docker configs / secrets


YuryHrytsuk:
```[tasklist]
### Requirements
- [x] Config updates shall propagate to the service automatically
  - [ ] both in CI Pipelines and when done manually
```

## Related issues
- https://github.com/ITISFoundation/osparc-ops-environments/issues/934
- https://github.com/ITISFoundation/osparc-ops-environments/issues/218
- https://github.com/docker/compose/issues/12504
- https://github.com/moby/moby/issues/35803

## Useful articles
- https://blog.sunekeller.dk/2019/01/docker-stack-deploy-update-configs/
- https://blog.viktoradam.net/2018/02/28/swarm-secrets-made-easy/
- https://github.com/neuroforgede/docker-stack-deploy
- [external secrets with name?](https://github.com/docker/compose/issues/5409)

## Useful scripts
- docker config decode https://gist.github.com/letsjustfixit/eb55d2cc12fcd34eae08b07c5eb8a93c

YuryHrytsuk:
We use external secrets for:
* portainer
* [local] traefik
* minio

It looks like docker config / secret name solution does not apply when `external` is `true`

PR #952, opened 2025-01-29
Introduce rolling docker config / secret update concept :tada: :rocket: 

YuryHrytsuk:
## What do these changes do?
* Add workflow to enable rolling (on-the-fly) docker config / secret updates.
* Provide an example on how to use this workflow via `appmotion gateway` api env config

## Details

- In order to guarantee rolling updates, add a hash suffix to the config name.
- This can also apply to external config / secrets via the same name-change mechanism.
- With this mechanism in place, it is important to clean up old docker config versions. For this reason, special makefile targets were introduced
- Find many useful articles, docs and resources in the description of  https://github.com/ITISFoundation/osparc-ops-environments/issues/949

## Related issue/s
* closes https://github.com/ITISFoundation/osparc-ops-environments/issues/949

## Checklist
- [X] I tested and it works


scripts/common.Makefile
```diff
@@ -207,9 +207,39 @@ clean-default: .check_clean ## Cleans all outputs
...
+	@# [docker config] versions are kept. This target removes them
+	@# https://github.com/docker/cli/issues/203
+	@#
+	@# This should be run before stack update in order to
+	@# keep previous config version for potential rollback
+	@#
+	@# This will not clean "external" configs. To achieve this extend
+	@# this target in related Makefiles.
+	@#
+	@# Long live Kubernetes ConfigMaps!
```
mrnicegyu11, line 228: <3

mrnicegyu11 (approved):
sieht gut aus, very surgical but this could work I see no issues with the code design. Good luck!

sanderegg (approved):
so you will update them on every release?

YuryHrytsuk:
> so you will update them on every release?

only on file content change

PR #950, opened 2025-01-27
Add fallback nginx to serve "something went wrong"

YuryHrytsuk:
## Related issue/s
- https://github.com/ITISFoundation/osparc-ops-environments/issues/218
- blocked by https://github.com/ITISFoundation/osparc-ops-environments/issues/949

## Checklist
- [ ] I tested and it works

Issue #218, opened 2023-06-26
EPIC: Refactor maintenance page related traefik routing


mrnicegyu11:
The maintenance page should always run as a kind of fallback/404 as a last priority. We tried running an oSparc with the maintenance page up with prio=1 (lowest prio), but the e2e/p2e showed problems. I closer look at all traefik routes and a fine-tuning are required to make this work properly.

YuryHrytsuk:
@mrnicegyu11 Do we really want to show `maintenance page` in case `webserver` is down? I think that we better show maintenance page when we really do maintenance in which case we can always set up it explicitly.  In case we show `maintenance page` every time the `webserver` cannot be reached. It can make users think that all fine and the maintenance is happening while we are actually have no clue that something is broken?

pcrespov:
The public-api is a machine-to-machine API. A fallback for maintenance should respond with [status 503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) 

![Image](https://github.com/user-attachments/assets/f89147c3-2d9e-4b42-908b-3d3a16b95249)

-  [status 503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) 
-  response JSON such as
```json
{
  "errors": [ "Under maintenance ... (human readable message)" ]
}
```
- headers attributes (with [`Retry` being optional](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After))
```http
content-type: application/json 
Retry-After: <http-date>
Retry-After: <delay-seconds>
```

YuryHrytsuk:
## Sustainable automated [docker] configs update process
- https://blog.sunekeller.dk/2019/01/docker-stack-deploy-update-configs/
- https://blog.viktoradam.net/2018/02/28/swarm-secrets-made-easy/

YuryHrytsuk:
After discussion with @sanderegg, we decided that this shall probably live in `osparc-simcore` repository (I am not sure why exactly so). To keep fallback service with main services in sync, we need to introduce ENV VARs for shared traefik configuration. In simcore we can also add integration tests to test fallback logic. In `ops-env` we will need to override configs definitions to ensure smooth config updates.

This is taking way more than expected. I will probably pause this for now.

YuryHrytsuk:
## Research: what happens if docker / traefik healthcheck fails

If traefik healthcheck fails for all replicas: 503 `no available server` is returned

PR #953, opened 2025-01-30
Update traefik router hardcoded priorities

YuryHrytsuk:
## What do these changes do?
* Update traefik router hardcoded priorities according to the rules below
  - 1 --> 3
  - 2 --> 6
  -  3 --> 9
  - 10 --> 12
* Remove unnecessary priority from `adminpanels` and `registry` service

This PR introduces some space in-between the routers which we can use to flexibly add new routers. It shouldn't break existing priority hierarchy since numbers <13 are still below any other non-hardcoded priorities.

Remarks for future:
* Whenever introducing some priorities to traefik routers (or other similar cases) one better to make them spacious to retain flexibility. This case is a good example why "spacious" approach makes sense

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/218

## Related PR/s
* https://github.com/ITISFoundation/osparc-simcore/pull/7141
* https://github.com/ITISFoundation/osparc-ops-environments/pull/950

## Checklist
- [ ] I tested and it works

### DIFF
--- a/services/admin-panels/docker-compose.yml.j2
+++ b/services/admin-panels/docker-compose.yml.j2
@@ -89,7 +89,7 @@ services:
         - traefik.http.services.adminpanels.loadbalancer.server.port=8888
         - traefik.http.routers.adminpanels.rule=Host(`${ADMINPANELS_DOMAIN}`)
         - traefik.http.routers.adminpanels.entrypoints=https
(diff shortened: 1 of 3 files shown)


mrnicegyu11:
I agree w.r.t. making the space of traefik priorities spacious, look good from my side

PR #967, opened 2025-02-05
💄 minor: Change DNS Server to Quad9

mrnicegyu11:
## What do these changes do?
Minor: Change DNS Server to Quad9

did this on the fly with "replace all"

Why Quad9?
- It's swiss
- It's privacy focused
- It's not google
## Checklist
- [ ] I tested and it works


services/admin-panels/docker-compose.aws.yml
```diff
@@ -2,4 +2,4 @@ version: '3.7'
 services:
   adminpanels:
     dns: # Add this always for AWS, otherwise we get "No such image: " for docker services
-      8.8.8.8
+      9.9.9.9
```
YuryHrytsuk, line 5: I wonder why we need this at all

YuryHrytsuk (approved):

services/admin-panels/docker-compose.aws.yml
mrnicegyu11: good point, but I still remember the mentioned issue, this thing was really the fix. This was from before LLMs got big so probably there is some stackoverflow post that explains why this fixes which bug

PR #1005, opened 2025-03-31
2025/add/graylog alerts

mrnicegyu11:
## What do these changes do?
@pcrespov This tries to address https://github.com/ITISFoundation/osparc-simcore/issues/7391 by making the graylog alert functionality working again. It also removes alerts that do not make sense anymore.
## Related issue/s
https://github.com/ITISFoundation/osparc-simcore/issues/7391
## Checklist
- [x] I tested and it works


services/graylog/scripts/alerts.template.yaml
```diff
@@ -432,6 +125,7 @@
     query_parameters: []
     search_within_ms: 86400000
     execute_every_ms: 86400000
+    event_limit: 1
```
YuryHrytsuk, line 128: Q: how does it work exactly? And when does limit get reset (to 0) so it can be triggered again?

services/graylog/scripts/configure.py
```diff
@@ -139,17 +145,96 @@ def configure_email_notifications(_session, _headers):
...
         noti
         for noti in _r.json()["notifications"]
-        if noti["title"] == "Graylog " + env.str("MACHINE_FQDN") + " mail notification"
+        if noti["title"] == "Graylog " + MACHINE_FQDN + " mail notification"
     ][0]["id"]
 
     return _mail_notification_id
 
 
+def configure_slack_notification_channel(_session, _hed) -> str:
```
YuryHrytsuk, line 154: Is there any test / way to verify that these notifications are configured properly and reach the final destination? 

(to avoid alerts that silently fail to reach slack)

YuryHrytsuk (approved):
Thanks!

services/graylog/scripts/configure.py
```diff
@@ -233,11 +312,16 @@ def configure_log_retention(_session, _headers):
         print("Error setting up graylog syslog capturing.")
 
 
-def configure_alerts(_mail_notification_id, _session, _headers):
+def configure_alerts(
```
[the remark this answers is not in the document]
mrnicegyu11: Since the alerts are setup in an idempotent way, I don't think this is necessary. This is the beauty of idempotent pipelines: Since the alerts are always deleted and then pushed according to the specs in the git-repo, the state of the git-repo is always enforced. If the deploy-ops pipeline fails, this highlights issues with the alerts

services/graylog/scripts/alerts.template.yaml
mrnicegyu11: I don't actually know tbh, this is a new required field, setting it to 1 gives the "normal" behavior, see: https://github.com/Graylog2/graylog2-server/issues/20785

services/graylog/scripts/configure.py
mrnicegyu11: I am not sure this can be done in a programmatic way in a satisfactory way. However, the graylog dashboards to the best of my knowledge indicate if alert channels cannot be reached. If you have a usggestion how to test this please let me know :--) i am open to it!

services/graylog/data/contentpacks/osparc-custom-content-pack-v2.json
```diff
@@ -623,7 +623,7 @@
             "configuration": {
               "grok_pattern": {
                 "@type": "string",
-                "@value": "log_level=%{WORD:log_level} \\| log_timestamp=%{TIMESTAMP_ISO8601:log_timestamp} \\| log_source=%{DATA:log_source} \\| (log_uid=%{WORD:log_uid} \\| )?log_msg=%{GREEDYDATA:log_msg}"
+                "@value": "log_level=%{WORD:log_level} \\| log_timestamp=%{TIMESTAMP_ISO8601:log_timestamp} \\| log_source=%{DATA:log_source} \\| (log_uid=%{WORD:log_uid} \\| (log_oec=%{WORD:log_oec} \\| )?log_msg=%{GREEDYDATA:log_msg}"
```
pcrespov, line 626: i thought you did this in another PR this morning? :-)

services/graylog/scripts/configure.py
```diff
@@ -139,17 +145,96 @@ def configure_email_notifications(_session, _headers):
...
 
     return _mail_notification_id
 
 
+def configure_slack_notification_channel(_session, _hed) -> str:
+    # Configure sending Slack notifications
+    if GRAYLOG_SLACK_WEBHOOK_URL != "":
+        assert GRAYLOG_SLACK_WEBHOOK_CHANNEL
+        assert GRAYLOG_SLACK_WEBHOOK_ICON_URL
+        print(
```
pcrespov, line 159: TIP: you should always use `logging` instead of `print` to log information.

services/graylog/scripts/configure.py
```diff
@@ -139,17 +145,96 @@ def configure_email_notifications(_session, _headers):
...
+                    == "Graylog " + MACHINE_FQDN + " Slack notification"
+                ][0]["id"]
+                return _slack_notification_id
+            print(
+                "Error while adding the Slack Notification. Status code of the request : "
+                + str(r.status_code)
+                + " "
+                + r.text
+            )
+            sys.exit(1)
```
pcrespov, line 223: check https://docs.python.org/3/library/os.html#os.EX_OK for all these error codes

services/graylog/scripts/configure.py
```diff
@@ -233,11 +312,16 @@ def configure_log_retention(_session, _headers):
         print("Error setting up graylog syslog capturing.")
 
 
-def configure_alerts(_mail_notification_id, _session, _headers):
+def configure_alerts(
+    _session,
```
pcrespov, line 316: I would annotate all your inputs/outputs so that static tools can assist you finding issues  ... specially since you do not have tests. a

matusdrobuliak66 (approved):

services/graylog/data/contentpacks/osparc-custom-content-pack-v2.json
mrnicegyu11: i did, the PRs are chained ;) this will only be merged after the other

services/graylog/scripts/configure.py
mrnicegyu11: thx, I changed it to this

services/graylog/scripts/configure.py
mrnicegyu11: will do, thx

services/graylog/scripts/configure.py
mrnicegyu11: you are right, but this specific enhancement will have to come another time, please feel free to re-raise this ;)

Issue #984, opened 2025-03-11
Implement rolling config / secret updates in existing services


YuryHrytsuk:
## TODO

- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1010
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1013
- [X] https://github.com/ITISFoundation/osparc-ops-environments/pull/1015
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1012
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1011
- [x] ~~registry pull through cache~~ no configs / secrets used
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1018
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1019
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1020
- [x] https://github.com/ITISFoundation/osparc-ops-environments/pull/1022
  - [x] alertmanager
  - [x] node exporter
  - [x] pgsql_query
  - [x] prometheus
  - [x] smokeping
  - [x] tempo
- [x] ~~portainer~~ password secret changes [do not come into affect](https://github.com/ITISFoundation/osparc-ops-environments/pull/1024) --> no sense
- [x] make sure all docker-compose.yml targets explicitly define dynamically generated config files

## Related PRs
- https://github.com/ITISFoundation/osparc-ops-environments/pull/952

## Related Issues
- https://github.com/ITISFoundation/osparc-ops-environments/issues/934

PR #1013, opened 2025-04-04
Traefik rolling config updates

YuryHrytsuk:
Bonus:
* Surround traefik rules in '' to avoid bash execution \`...\`
* Set conservative and slow update_config to make sure only 1 containers at a time is updated (with a noticeable delay in between)

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/984

## Checklist
- [x] I tested and it works


services/traefik/Makefile
```diff
@@ -9,33 +9,32 @@ REPO_BASE_DIR := $(shell git rev-parse --show-toplevel)
 # TARGETS --------------------------------------------------
 include ${REPO_BASE_DIR}/scripts/common.Makefile
 
-
 .PHONY: up-local
-up-local: .init .create_secrets .env ${TEMP_COMPOSE}-local ## Deploys the stack using provided certificates
+up-local: .init .create_secrets ${TEMP_COMPOSE}-local prune-docker-stack-configs ## Deploys the stack using provided certificates
```
mrnicegyu11, line 13: does `prune-docker-stack-configs` depend on `.env`? I think the `.env` target still needs to be called, the .env built.

mrnicegyu11 (approved):
very nice thx

services/traefik/Makefile
YuryHrytsuk: `prune-docker-stack-configs` does not depend on `.env`

YuryHrytsuk: > I think the .env target 

Why?

mrnicegyu11: After talking about I understand this change, it is ok :--) thx

YuryHrytsuk: cleared in person. `.env` file is already defined in compose deps and here is not necessary

PR #1037, opened 2025-05-06
Add: graylog system notifications cleaner script

mrnicegyu11:
## What do these changes do?
If a graylog alert fires more than >1000 times within the check-period, the `event_limit` is reached and a graylog system notification is created. Too many graylog system notifications can slow down the graylog frontend significantly. This script facilitates the cleanup of these system notifications
## Checklist
- [x] I tested and it works


YuryHrytsuk (commented):
I am not in favor of this way. 

We add more code to fix a problem but problem is sitll here. This increases our technical debt but does not solve the problem (alleviates it) + adds extra manual step required to keep our platforms running.

Having said that, I would wait for @pcrespov to come back and discuss it. In the meantime we can simple remove these events from Graylog since anyway (it seems) nobody uses them.

@mrnicegyu11 what do you think

mrnicegyu11:
Yes I agree that this is on the "client" so the people that create the alerts - at least for now until we maybe have `fluentd+loki` maybe. 

I dont think we must merge this, I dont think we should run this regularly. I saw this more like the script that we have to delete pupeteer's old projects, something to run manually in case stuff piles up. So this code can also "live" in this PR and if needed it can be dug out :--) The proper way is fixing the error.


Let me know how you prefer to have it, a closed PR or merge for now and delete the script later, I am fine with both :--) @YuryHrytsuk

PR #1046, opened 2025-05-13
traefik v3.4 deprecated `traefik.docker.` in favor of `traefik.swarm.` labels

sanderegg:
## What do these changes do?
should fix zillions of `WRN Labels traefik.docker.* for Swarm provider are deprecated. Please use traefik.swarm.* labels instead` in traefik logs
## Related issue/s
- relates to https://github.com/ITISFoundation/private-issues/issues/79

Please @YuryHrytsuk @mrnicegyu11 can you double check that this does not go in staging or prod at the wrong time? shall I create multiple PRs? how does this one work?

## Checklist
- [ ] I tested and it works

### DIFF
--- a/services/admin-panels/docker-compose.yml.j2
+++ b/services/admin-panels/docker-compose.yml.j2
@@ -87,7 +87,7 @@ services:
     deploy:
       labels:
         - traefik.enable=true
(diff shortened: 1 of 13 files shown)


mrnicegyu11 (approved):
thanks a lot!

Issue #1048, opened 2025-05-15
Add detailed rabbitmq monitoring


YuryHrytsuk:
## Tasks
- [x] [master](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1402)
- [x] [staging](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1402)
- [x] [prod](https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1403)

## Useful links
* RabbitMQ Prometheus & Grafana docs
  - https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbitmq_prometheus/docker/docker-compose-overview.yml 
* https://www.robustperception.io/getting-cloudwatch-data-into-prometheus/

PR #1050, opened 2025-05-16
Add cloudwatch grafana datasource

YuryHrytsuk:
## What do these changes do?
This lets us build dashboards using cloudwatch metrics (aws managed
resources like RDS or MQ store their prometheus-like metrics there)

In future we can add dashboards for Postgres / Redis. Most likely we can reuse existing dashboards created by other people (for sure for RDS)

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/1048

## Checklist
- [ ] I tested and it works

### DIFF
--- a/services/monitoring/grafana/terraform/datasources.tf
+++ b/services/monitoring/grafana/terraform/datasources.tf
@@ -23,3 +23,18 @@ resource "grafana_data_source" "tempo" {
   basic_auth_enabled = false
   is_default         = false
 }
+
+resource "grafana_data_source" "cloudwatch" {
+  type = "cloudwatch"
+  name = "cloudwatch"
+
+  json_data_encoded = jsonencode({
+    defaultRegion = var.AWS_DEFAULT_REGION
+    authType      = "keys"
+  })
+
+  secure_json_data_encoded = jsonencode({
+    accessKey = var.AWS_GRAFANA_CLOUDWATCH_DATASOURCE_USER_ACCESS_KEY
+    secretKey = var.AWS_GRAFANA_CLOUDWATCH_DATASOURCE_USER_SECRET_KEY
+  })
+}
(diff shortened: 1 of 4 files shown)


sanderegg (approved):
thanks! that is very nice!

matusdrobuliak66 (approved):
Thanks!

services/monitoring/grafana/terraform/datasources.tf
```diff
@@ -23,3 +23,19 @@ resource "grafana_data_source" "tempo" {
   basic_auth_enabled = false
   is_default         = false
 }
+
+resource "grafana_data_source" "cloudwatch" {
```
mrnicegyu11, line 27: We can also add datasources the way we add dashbaords, i.e. use folders in `ospar-ops-deployment-configuration`, scrape/walk those fodlers for files and them add those files found.

This way we have "common" dashboards (in `deployments/shared/`) and per-deployment dashboards. The same could work for data_sources. i only didnt implement this to not blow up the grafana-tf PRs, that were anyway already painful enough.

Not sure how your solution would look like or react in an on-prem deployment, but in case you tested it (e.g. osparc.local) and it's ok, I am also fine with merging as is.

mrnicegyu11 (approved):
Check my comment, very nice push thank you! A thought: In the past we had postgresDB datasources, but they were unmaintained and unmaintainable with the old script-based provisioning system for grafana, or before that when we had nothing just chaos :D  maybe now we can bring them back?

services/monitoring/grafana/terraform/datasources.tf
YuryHrytsuk: Good find. I added a flag AWS / NON-AWS to be used to disable / enable resource creation by terraform 

@mrnicegyu11

mrnicegyu11 (approved):
thx that works!

PR #1055, opened 2025-05-20
Simcore helm charts: use external db envs

YuryHrytsuk:
## What do these changes do?
Since database sometimes run inside docker swarm, default envs like POSTGRES_HOST will contain internal swarm DNS not resolvable from the outside. For that we introduce new env, that is available withing networks (but not publicly available)

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/1052

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1406

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
very reasonable :) merci

PR #1065, opened 2025-05-28
Fix director metrics port

YuryHrytsuk:
## What do these changes do?
This shall fix issue of lacking director metrics in grafana (prometheus scraping port is wrong)

## Checklist
- [x] I tested and it works


bisgaard-itis (approved):
Thanks a lot for the quick fix

PR #1070, opened 2025-06-05
Traefik: use compose verson 3.12 :warning: 

YuryHrytsuk:
## What do these changes do?
Since introduction of traefik healthcheck we need settings that are not supported by 3.7

Error message:
```bash
services.traefik.healthcheck Additional property start_interval is not allowed
```

Bonus:
* Remove unused dns and http compose files

## Related PR/s
* https://github.com/ITISFoundation/osparc-ops-environments/pull/1068

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/traefik/docker-compose.master.yml
+++ b/services/traefik/docker-compose.master.yml
@@ -1,5 +1,3 @@
-version: "3.12"
-
 services:
     traefik:
         command:
--- a/services/traefik/docker-compose.aws.yml
+++ b/services/traefik/docker-compose.aws.yml
@@ -1,5 +1,3 @@
-version: "3.12"
-
 services:
   traefik:
     command:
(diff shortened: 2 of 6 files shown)


services/traefik/docker-compose.master.yml
```diff
@@ -1,4 +1,5 @@
-version: "3.7"
+version: "3.12"
```
sanderegg, line 1: so just for info, these version top field should not be necessary anymore:
https://docs.docker.com/reference/compose-file/version-and-name/

sanderegg (commented):
thanks a lot!
Please check this https://docs.docker.com/reference/compose-file/version-and-name/ for later. the version top field is obsolete.

sanderegg (approved):
thanks!

Issue #1080, opened 2025-06-18
Fix / Add postgres grafana dashboard


YuryHrytsuk:
Implementation https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1443

To do list (deploy to):
- [x] master https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1443
- [x] stag https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1444
  - [x] + missing https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1446
- [x] prod https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1445
  - [ ] configure iam user for grafana's cloudwatch datasource https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1447

PR #1081, opened 2025-06-18
Update postgres exporter

YuryHrytsuk:
## What do these changes do?

Make sure we use recent version that has export all fields used by existing community grafana Postgres dashboards

## Related issue/s
* https://github.com/ITISFoundation/osparc-ops-environments/issues/1080

## Checklist
- [ ] I tested and it works


mrnicegyu11 (approved):
thx

PR #1118, opened 2025-07-08
Refactor: ops-traefik configuration (2nd attempt)

mrnicegyu11:
## What do these changes do?
Re-iteration of https://github.com/mrnicegyu11/osparc-ops-environments/pull/1

Emerged as a needed refactoring to facilitate https://github.com/ITISFoundation/osparc-ops-environments/pull/1041#discussion_r2084115662

Sets the traefik-config using env-vars instead of CLI arguments, avoids the need for per-deployment-target docker-compose files (except `osparc.local`)

 
 ##### Bonus:
- Removes env-var `REDIS_EXTERNAL_HOST`
- Removes docker compose yaml version specification
 - (in osparc-config) adds `osparc.local/cert_domains.json`
 - Bugfix: sets `TRACING_OPENTELEMETRY_COLLECTOR_SAMPLING_PERCENTAGE` to 100 everywhere
## Related PR/s
https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1482

## Checklist
- [x] I tested and it works


charts/simcore-charts/resource-usage-tracker/values.yaml.gotmpl
```diff
@@ -135,9 +135,9 @@ env:
   - name: REDIS_USER
     value: {{ env "REDIS_USER" }}
   - name: REDIS_HOST
-    value: {{ requiredEnv "REDIS_EXTERNAL_HOST" }}
```
YuryHrytsuk, line 138: This env is necessary because RUT runs on Kubernetes and cannot access redis via Docker Swarm DNS. While other simcore services run inside docker swarm, they can rely on `REDIS_HOST` which is Docker Swarm DNS of `redis`. 

RUT in the meantime, uses DNS that is available outside of Docker Swarm

services/simcore/docker-compose.deploy.master.yml
```diff
@@ -35,7 +35,7 @@ services:
       - "--entryPoints.http.transport.respondingTimeouts.writeTimeout=21600s" #6h, for https://github.com/traefik/traefik/issues/10805
       - "--entryPoints.http.transport.respondingTimeouts.readTimeout=21600s" #6h, for https://github.com/traefik/traefik/issues/10805
       - "--entryPoints.postgres.address=:5432"
-      - "--entryPoints.redis.address=:${REDIS_EXTERNAL_PORT}"
+      - "--entryPoints.redis.address=:5672"
```
YuryHrytsuk, line 38: If we keep redis external port env, I would use the ENV

services/traefik/docker-compose.yml.j2
```diff
@@ -57,18 +18,25 @@ services:
...
+      {% if OPS_TRAEFIK_EXPOSE_RABBITMQ|lower == "true" %}
+      - target: 5672
+        published: 5672
         mode: host
+      {% endif %}
     configs:
       - source: traefik_dynamic_config.yml
         target: /etc/traefik/dynamic_conf.yml
+    env_file:
+      - .env
```
YuryHrytsuk, line 35: Question: does docker swarm requires `.env` file to be always present or it can be removed after being deployed? 

Since in CI we always clean up files, `.env` will be removed after deploy

mrnicegyu11: Good point, I am 99% sure this is only needed at deploy time and can be removed, but I will double-check it

charts/simcore-charts/resource-usage-tracker/values.yaml.gotmpl
mrnicegyu11: alright understood thx, I will add it again :--)

mrnicegyu11: done

services/simcore/docker-compose.deploy.master.yml
mrnicegyu11: the env has been added again

mrnicegyu11:
this was successfully tested on `osparc.local`

Issue #1089, opened 2025-06-25
sim4life and nih prod: vendors image cannot be pulled by portainer (--> portainer needs to be authenticated via registry)


YuryHrytsuk:

YuryHrytsuk:
Issues from portainer:
* https://github.com/portainer/portainer/issues/2341
* https://github.com/portainer/portainer/issues/1313

PR #1125, opened 2025-07-15
Add portainer registry configuration 🚨

YuryHrytsuk:
## What do these changes do?
Creates a single authneticated dockerhub registry in Portainer. With this registry, private docker images can be pulled.

Bonus:
* Add reusable $(WAIT_FOR_IT) target that calls https://github.com/wait4x/wait4x and supports many protocols
* Extend `install` target (requirements.txt file's path can be configured)

## Manual testing
* Check that portainer can pull private images on PROD deployments

## Related issue/s
* closes https://github.com/ITISFoundation/osparc-ops-environments/issues/1089

## Related PR/s
* Configuration https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/1489

## Checklist
- [x] I tested and it works


scripts/common.Makefile
```diff
@@ -362,3 +367,16 @@ define jinja
...
 endif
+
+#
+# wait-fot-it functionality
+#
+
+WAIT_FOR_IT := $(REPO_BASE_DIR)/scripts/wait4x
+
+# https://github.com/wait4x/wait4x
+$(WAIT_FOR_IT):
```
mrnicegyu11, line 378: We are doing Makefiles on steroids here, shouldn't this correctly be PHONY as it should always run?

services/portainer/scripts/configure_portainer_registry.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        raise
+
+    return response.json()
+
+
+def main():
+    logger.info("Configuring Portainer registries...")
+
+    portainer_username = os.environ["SERVICES_USER"]
```
mrnicegyu11, line 100: Minor: As Pedro has converted me into a believer, I will also try to missionize you: For nice python CLI interfaces, consider using `typer`. Usually the LLM can simply add it if you ask to rewrite the script using typer ;)

scripts/common.Makefile
YuryHrytsuk: wrt steroids. I partially agree. Although I believe `wait_for_it` installation is not complicated, it is still some extra logic and dependency coming from the internet. 

I also believe that the root for these steroids is the fact that we do have this "wait-for-it" logic in makefile. We can use our own code (e.g. write a simple python script) but it won't change much

mrnicegyu11 (approved):
all good, approved!

PR #1131, opened 2025-07-17
Fix portainer makefile

YuryHrytsuk:
## What do these changes do?
Spell correctly no print directory option of `make`

## Related PR/s
* Bug introduced in https://github.com/ITISFoundation/osparc-ops-environments/pull/1125

## Checklist
- [x] I tested and it works


mrnicegyu11 (commented):
the best!

PR #1129, opened 2025-07-16
update grok pattern

bisgaard-itis:
## What do these changes do?
- Since the async logger went in (https://github.com/ITISFoundation/osparc-simcore/pull/8064) we cannot search for fields in Graylog (e.g. the trace id) because one of the GROK patterns were wrong. There were two symptoms I observed when searching for fields in Graylog:
  - Some fields could not be searched (e.g. the `log_trace_id`)
  - Some fields were parsed incorrectly (e.g. the `log_span_id`).
- This PR fixes that.
## Related PR/s
- https://github.com/ITISFoundation/osparc-simcore/pull/8064
- https://github.com/ITISFoundation/osparc-simcore/pull/6857
## Checklist
- [x] I tested and it works

PR #1133, opened 2025-07-17
Revert "update grok pattern (#1129)"

bisgaard-itis:
This reverts commit 1865d841072bc94fdb60780e4ef9c1a8b9a4b454.

## Checklist
- [ ] I tested and it works


YuryHrytsuk:
This revert was not enough to fix the problem. `cd services/graylog && make configure` kept failing.

error log
```
Caused by: com.mongodb.DuplicateKeyException: Write failed with error code 11000 and error message 'Write operation error on server mongo:27017. Write error: WriteError{code=11000, message='E11000 duplicate key error collection: graylog.pipeline_processor_pipelines index: title_1 dup key: { title: "osparc-simcore replace pipeline" }', details={}}.'
```

Manually removing this pipeline via Graylog GUI (System / Content Packs --> Pipelines --> Choose Pipeline & Delete it) fixed the issue.

It is also important to understand what happens to content pack versions. This revert a version to previous one (but both previous version and current version being reverted are already present in Graylog). TODO

PR #1153, opened 2025-07-28
Kubernetes: refactor local cluster deployment

YuryHrytsuk:
## What do these changes do?
Changes
* Use calico CNI for networking as we use it in master deployments
* Move local-reated targets to a seperate Makefile (keep clean main Makefile used in master/stag/prod)
* Add delete (local) cluster target

## Checklist
- [x] I tested and it works


sanderegg (approved):
💋

mrnicegyu11 (approved):

PR #1171, opened 2025-08-07
fix graylog grok pattern

matusdrobuliak66:
## What do these changes do?
fix graylog grok pattern
## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/pull/8191
## Checklist
- [ ] I tested and it works


bisgaard-itis (approved):
Thanks a lot for the effort 👍🏻

Issue #1167, opened 2025-08-05
Change rollback policy on Simcore stack


matusdrobuliak66:
Remove rollback policy from STO workers in the simcore stack, or from the entire simcore stack?

**Why?**
We had two incidents where the STO worker performed a rollback, which caused side effects such as not shutting down sidecars and services properly. This led to financial losses.

YuryHrytsuk:
Partially related https://github.com/ITISFoundation/osparc-ops-environments/issues/940

PR #1174, opened 2025-08-08
Simcore: remove rollback update policy

YuryHrytsuk:
## What do these changes do?
simcore service should all be using the same version. Version mismatches are not expected (developers do not take backwarn compatibility between simcore service versions into account). Also simcore services using database expect a migraton version to match (directly related to version).

FYI: @pcrespov @sanderegg @GitHK @bisgaard-itis 

## Related issue/s
* closes https://github.com/ITISFoundation/osparc-ops-environments/issues/1167

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/simcore/docker-compose.yml.j2
+++ b/services/simcore/docker-compose.yml.j2
@@ -346,7 +346,7 @@ services:
       update_config:
         parallelism: 2
         order: start-first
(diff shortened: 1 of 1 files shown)


mrnicegyu11 (approved):
nice, glad to have this merci!

PR #1175, opened 2025-08-08
Kubernetes: add `cert-manager` network policy

YuryHrytsuk:
## Checklist
- [ ] I tested and it works


charts/cert-manager/values.common.yaml.gotmpl
```diff
@@ -1,6 +1,10 @@
-crds:
-  enabled: true
-  keep: true
+cert-manager:
```
mrnicegyu11, line 1: re changes in this file: are they due to  new syntax requirements?

mrnicegyu11 (approved):
lgtm thx!

charts/cert-manager/values.common.yaml.gotmpl
YuryHrytsuk: original cert-manager was made a subchart. So all values we want to pass to original cert-manager chart now need to be inside key `cert-manager`. 

See more https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

charts/cert-manager/values.selfsigned.yaml.gotmpl
```diff
@@ -1,56 +1,57 @@
...
-    commonName: local-ca
-    subject:
-      organizations:
-        - Z43
-    secretName: local-ca-secret
-    privateKey:
-      algorithm: ECDSA
-      size: 256
-    issuerRef:
+cert-manager:
```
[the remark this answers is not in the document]
YuryHrytsuk: No syntax change. `cert-manager` is now a subchart, so to pass values we need to add key `cert-manager`

https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

PR #1190, opened 2025-08-27
Automatical lokal loki bucket creation

mrnicegyu11:
## What do these changes do?
Auto-Magically create Loki S3 bucket on local deployment

## Checklist
- [x] I tested and it works


services/logging/Makefile
```diff
@@ -7,11 +7,24 @@ STACK_NAME = $(notdir $(shell pwd))
 TEMP_COMPOSE=.stack.${STACK_NAME}.yaml
 REPO_BASE_DIR := $(shell git rev-parse --show-toplevel)
 
+define create-s3-bucket
```
YuryHrytsuk, line 10: create-local-s3-bucket?

mrnicegyu11: we have the existing name already in the registry stack so ill keep it around for now ;) but yours would be better in theory

Issue #1186, opened 2025-08-25
FluentD is overwriting the source field


sanderegg:
<img width="3608" height="451" alt="Image" src="https://github.com/user-attachments/assets/672996de-4c4c-4bc3-9c47-cf4a3c331f18" />

mrnicegyu11:
To the best of my knowledge fixed by @sanderegg  in https://github.com/ITISFoundation/osparc-ops-environments/pull/1189

sanderegg:
still a problem

Issue #1185, opened 2025-08-25
fluentd fails to flush a buffer in Loki


YuryHrytsuk:
It happens in both master

```
2025-08-25 12:45:19 +0000 [warn]: #0 failed to flush the buffer. retry_times=1 next_retry_time=2025-08-25 12:45:22 +0000 chunk="63d2f9679835a43828a1ab1c6cff93c1" error_class=Fluent::Plugin::LokiOutput::LogPostError error="429 Too Many Requests Ingestion rate limit exceeded for user fake (limit: 4194304 bytes/sec) while attempting to ingest '11447' lines totaling '8166288' bytes, reduce log volume or contact your Loki administrator to see if the limit can be increased\n"
```

PR #1206, opened 2025-09-10
✨ Add: Fluentbit log aggregator (removes fluentd)

mrnicegyu11:
## What do these changes do?
This is an improvement on the first iteration of having a logging-aggregator pass the logs of docker to both graylog and loki. First, we tried with `fluentd`, which would ingest the docker logs in the `fluentd` format.

It turns out that the fluentd logging format used by dockerd does not contain the hostname of the machine where the docker container that logs runs. This renders it unusable for the use-case of identifying issues on autoscaled EC2 machines. It furthermore turned out that `fluentd` was a pain to configure, and was not very performant (log dropping due to slowness, CPU hungryness).

This change substitutes `fluentd` for the more modern `fluentbit`. Fluentbit is more performant and versatile. It is specifically able to *ingest* GELF formatted logs and pass them unchanged to graylog, which is desirable for us to maintain backwards-compatibility and to assure that the hostname of the machines are added to the logs.

This change was successfully tested locally, but requires the linked osparc-infra PR to disable compression of the UDP GELF logs sent by dockerd, as `fluentbit` cannot decompress them. This change is also "nicer" in the sense that less custom code or docker images are needed, inly native functionality of `fluentbit` is used. `Loki` was also asserted to work fine with this.

## Related issue/s
https://github.com/ITISFoundation/osparc-ops-environments/issues/1185
https://github.com/ITISFoundation/osparc-ops-environments/issues/1186

## Related PR/s
Requires https://git.speag.com/oSparc/osparc-infra/-/merge_requests/310
## Checklist
- [ ] I tested and it works


mrnicegyu11:
This is closed in favor of vector-dev

PR #1243, opened 2025-10-27
Process logs before sending them to loki

bisgaard-itis:
## What do these changes do?
- Extract fields in Grok pattern so that they are more easy to query/filter in Loki.
- Only add Loki labels which have low cardinality (https://grafana.com/docs/loki/latest/get-started/labels/bp-labels/)

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/logging/loki.yaml
+++ b/services/logging/loki.yaml
@@ -43,6 +43,7 @@ compactor:
   retention_enabled: false
 
 limits_config:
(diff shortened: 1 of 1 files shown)


mrnicegyu11 (commented):
@bisgaard-itis please consider YH's comments, or at least the traefik one. Thanks a lot, merge at your own discretion please

PR #1261, opened 2025-11-14
Improve vector log processing

bisgaard-itis:
## What do these changes do?
Improve vector's processing of logs to improve user experience in Loki. More precisely, as discussed with @mrnicegyu11, this updates the vector log processing in the following ways:
- When ever a log has a well-defined `level` (as specified by the `log_level`), this is added as a label to the log message. Otherwise `level="UNKNOWN"` is set. This is done in order to avoid that Loki doesn't try to guess the log level (which doesn't seem to work very well for our logs).
- `\n` are resolved to display new lines in the log. This makes the reading of logs in loki much easier.
- If a `log_msg` is not found in the log (e.g. because it is a log line which is generated before the proper log formatting is setup inside our simcore services) then `log_msg=msg` in order to ensure that a message will be displayed in loki. 

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/logging/vector.yaml
+++ b/services/logging/vector.yaml
@@ -90,7 +90,7 @@ transforms:
       }
 
       # Add processing metadata
-      .processed_by = "vector-dev"
+      .processed_by = "vector"
 
 sinks:
   # Send to Loki over TCP


services/logging/vector.yaml
```diff
@@ -38,14 +40,26 @@ transforms:
           .log_trace_id = parsed_fields.log_trace_id
           .log_span_id = parsed_fields.log_span_id
           .log_msg = parsed_fields.log_msg
-        }
-        # traefik logs
-        traefik_fields, traefik_err = parse_regex(.message, r'time="(?P<log_timestamp>[^"]*)" level=(?P<log_level>\w+) msg="(?P<log_msg>.*)"')
-        if traefik_err == null {
+        } else if traefik_err == null {
```
YuryHrytsuk: Q: why we don't rely on `log_service` or some check like `if "traeifk" in container.name` which is more explicit, precise and self-explanatory?

PR #1263, opened 2025-11-17
Fix graylog log processing incompatibility

bisgaard-itis:
## What do these changes do?
- Vector stopped processing logs with the following error
```
2025-11-17T07:04:53.221877Z ERROR sink{component_kind="sink" component_id=graylog component_type=socket}: vector::internal_events::codecs: Failed serializing frame. error=LogEvent contains a value with an invalid type. field = "level" type = "string" expected type = "integer" error_code="encoder_serialize" error_type="encoder_failed" stage="sending" internal_log_rate_limit=true
```
- Apparently the issue is that graylog doesn't allow a field `.level` with a string value (it is required to be an int). 
- This PR fixes the issue by splitting the log processing in vector into two different stages: A base and one specialized for Loki.

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/logging/vector.yaml
+++ b/services/logging/vector.yaml
@@ -15,7 +15,7 @@ sources:
 
 transforms:
   # Process and enrich the logs
-  process_logs:
+  process_logs_base:
     type: remap
     inputs: ["docker_gelf"]
     source: |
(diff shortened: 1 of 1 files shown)


services/logging/vector.yaml
```diff
@@ -25,6 +25,41 @@ transforms:
...
+      } else {
+        .image_name = "unknown"
+      }
+
+      # Add processing metadata
+      .processed_by = "vector"
+
+  process_logs_loki:
+    type: "remap"
+    inputs: ["process_logs_base"]
```
mrnicegyu11: uh, cool stuff that you can chain them

PR #1321, opened 2026-01-12
Parse `log_trace_sampled` in vector-dev config (logging stack)

bisgaard-itis:
## What do these changes do?
This pull request updates the log parsing regex for Python service logs in the `vector.yaml.j2` configuration to extract an additional field, `log_trace_sampled`, from the log messages. This change improves the structured logging by capturing more trace-related information.

Log parsing improvements:

* Updated the regex pattern in the `transforms` section of `services/logging/vector.yaml.j2` to extract a new field, `log_trace_sampled`, from Python service logs.

## Related issue/s
- https://github.com/ITISFoundation/osparc-simcore/pull/8736

## Checklist
- [ ] I tested and it works
- There is a test on the osparc-simcore side which checks that the parsing with this regex pattern is correct.

### DIFF
--- a/services/logging/vector.yaml.j2
+++ b/services/logging/vector.yaml.j2
@@ -81,6 +81,7 @@ transforms:
           .log_oec = parsed_fields.log_oec
           .log_trace_id = parsed_fields.log_trace_id
           .log_span_id = parsed_fields.log_span_id
+          .log_trace_sampled = parsed_fields.log_trace_sampled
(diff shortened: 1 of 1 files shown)


bisgaard-itis:
> Thanks 🙏
> 
> Maybe title could be extended to be more clear at first sight (e.g. mentioning logging stack)

Thanks, good point. Done

mrnicegyu11 (approved):
Thx lgtm

PR #1342, opened 2026-02-05
Fix: Loki (Remove distroless image healthchecks)

mrnicegyu11:
## What do these changes do?
Since 3.5.8., the loki docker image is distroless, which means there is absolutely no shell included. Thus, no healthchecks can be performed as docker implictly runs healthcheck commands in `/bin/sh`.

See: https://grafana.com/docs/loki/latest/setup/upgrade/

image

image

The only way feasible in docker (add dependencies on top of the base image) is explicitly not recommended in prod.

The only way to get a healthcheck is a move to k8s is seems

### Bonus:
`stop-first` update policy hopefully makes the restarting of `loki` smoother, as currently it will hang forever since the old container is still exclusively bound to the host port.

## Related issue/s
https://git.speag.com/oSparc/osparc-infra/-/issues/93
## Checklist
- [ ] I tested and it works

### DIFF
--- a/certificates/Makefile
+++ b/certificates/Makefile
@@ -66,10 +66,16 @@ install-root-certificate: rootca.crt ## installs a certificate in the host syste
 		echo "Is the DOCKER service ready? press when ready" && read -n 1; \
 	    fi;\
 		echo "======================================";,\
-		sudo cp $< /etc/ca-certificates/trust-source/anchors/osparc.crt; \
-		sudo trust extract-compat &&                            \
-		echo "# restarting docker daemon" &&                      \
+		$(if $(IS_OSX),                                             \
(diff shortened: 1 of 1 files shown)


services/logging/docker-compose.yml.j2
```diff
@@ -182,15 +182,9 @@ services:
       update_config:
         parallelism: 1
         delay: 10s
-        order: start-first
+        order: stop-first # Due to host port binding
     networks:
       - monitoring
-    healthcheck:
```
YuryHrytsuk: Does it mean there is no healthcheck for this service now? Or is there a default healthcheck defined in Dockerfile? @mrnicegyu11

PR #1368, opened 2026-03-11
:bug: :alembic: Fix: Docker29 yaml parsing

mrnicegyu11:
## What do these changes do?
Fix: Docker29 yaml parsing, not accepting duplicate keys

Bonus: Add a check and a verbose clear error if this is the case somewhere
## Checklist
- [x] I tested and it works

### DIFF
--- a/certificates/Makefile
+++ b/certificates/Makefile
@@ -66,10 +66,16 @@ install-root-certificate: rootca.crt ## installs a certificate in the host syste
 		echo "Is the DOCKER service ready? press when ready" && read -n 1; \
 	    fi;\
 		echo "======================================";,\
-		sudo cp $< /etc/ca-certificates/trust-source/anchors/osparc.crt; \
-		sudo trust extract-compat &&                            \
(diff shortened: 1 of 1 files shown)


scripts/deployments/compose_stack_yml.bash
```diff
@@ -80,6 +80,36 @@ envsubst < .env.nosub > .env
 cp .env ..
 cp ../../docker-compose.yml ./docker-compose.simcore.yml
 
+# Docker29
+# Verify that docker-compose.deploy.yml defines replicas for every service
```
YuryHrytsuk: Does it mean that every service replicas has to be specified in 4 deploy files even when the number is the same?

YuryHrytsuk:
Summary

I think we can leave without any changes to creation of simcore stack file which is already unweild and hard to follow. 

We can add a pre-commit hook to ensure that replicas settings are either in compose.yml.j2 or in all deploy files (wrt simcore compose file we ask simcore developers to remove any replica settings from their compose file). This shall suffice

PR #1386, opened 2026-04-09
Fix: Remove gzip recompression from Docker registry router

GitHK:
### Problem

Docker image layers are already gzip-compressed (`application/vnd.docker.image.rootfs.diff.tar.gzip`). The registry router was applying `ops_gzip@swarm`, causing Traefik to recompress these binary blobs on every pull — wasting CPU and buffering hundreds of MB in RAM for zero compression benefit.

### Changes

**File:** docker-compose.yml.j2

1. **Removed `ops_gzip@swarm` from the registry middleware chain**
   - Before: `middlewares=ops_gzip@swarm, ops_auth@swarm`
   - After: `middlewares=ops_auth@swarm`

2. **Added `flushinterval=1ms` to the registry loadbalancer**
   - Forces Traefik to stream responses immediately to the Docker client instead of accumulating data, reducing latency on large blob transfers.

### What was NOT changed

- `ops_auth@swarm` remains — registry authentication is preserved.
- No buffering middleware was added — Traefik does not buffer by default, so explicitly disabling it is unnecessary.
- No `Accept-Encoding` header stripping — `ops_gzip` is per-router (not global at the entrypoint level), so simply removing it from the middleware chain is sufficient.
- The pull-through-cache service is unaffected — it binds directly to host port 5000 and does not go through Traefik.

### Expected impact

- Lower CPU usage on Traefik during image pulls
- Lower memory usage (no buffering of compressed blobs)
- Reduced pull latency due to streaming flush

## Related issue/s
- follow up to https://github.com/ITISFoundation/osparc-simcore/issues/8875#issuecomment-4212873859

## Checklist
- [ ] I tested and it works

### DIFF
--- a/services/registry/docker-compose.yml.j2
+++ b/services/registry/docker-compose.yml.j2
@@ -59,10 +59,14 @@ services:
         - traefik.swarm.network=${PUBLIC_NETWORK}
         # direct access through port
         - traefik.http.services.registry.loadbalancer.server.port=5000
+        # stream large blob transfers immediately instead of buffering
+        - traefik.http.services.registry.loadbalancer.responseforwarding.flushinterval=1ms
         - traefik.http.routers.registry.rule=Host(`${REGISTRY_DOMAIN}`)
         - traefik.http.routers.registry.entrypoints=https
         - traefik.http.routers.registry.tls=true
-        - traefik.http.routers.registry.middlewares=ops_gzip@swarm, ops_auth@swarm
+        # NOTE: ops_gzip removed — Docker image layers are already gzip-compressed;
+        # re-compressing wastes CPU and buffers large blobs in RAM
+        - traefik.http.routers.registry.middlewares=ops_auth@swarm
         - prometheus-job=registry
         - prometheus-port=5001
 


mrnicegyu11 (approved):
very very good. I have read a long time ago in some blogpost that gzipping middleware is oldschool stuff that with current computing/networking noone actually needs, it argued the overhead of zipping is a higher price to pay than the smaller request size. Sadly I dont have that blog post link anymore. The gzip middleware was already firmly in place when I joined the company. Maybe we should test removing it alltogether? @YuryHrytsuk thoughts?


Thanks a lot

GitHK:
> @GitHK are there any non-AI proofs that docker actually pulls already gzipped chunks?

@YuryHrytsuk As you asked I have ran some tests agains AWS master. This will fix the issue.

Notes on the screenshots below:
- The mostly `red` line is the test without the headers on the registry
- The mostly `blue` line is the test with the suggested headers added to the registry
- The `red` test was shorter then the `blue` I can't tell by how much
- you can tell the test is ongoing because CPU goes up and network usage goes up (less evident since it always goes up a bit)
- in the `blue` test RAN stays flat no more pressure from the registry, no more restarts


traefik_1
traefik_2
treafik_3

YuryHrytsuk:
> > @GitHK are there any non-AI proofs that docker actually pulls already gzipped chunks?
> 
> @YuryHrytsuk As you asked I have ran some tests agains AWS master. This will fix the issue.
> 
> Notes on the screenshots below:
> 
>     * The mostly `red` line is the test without the headers on the registry
> 
>     * The mostly `blue` line is the test with the suggested headers added to the registry
> 
>     * The `red` test was shorter then the `blue` I can't tell by how much
> 
>     * you can tell the test is ongoing because CPU goes up and network usage goes up (less evident since it always goes up a bit)
> 
>     * in the `blue` test RAN stays flat no more pressure from the registry, no more restarts
> 
> ...

Thanks for the test. 

> The mostly blue line is the test with the suggested headers added to the registry

I am a bit confused by the wording. The PR suggests to remove headers (not to add them). Could you elaborate what is blue / red line? Maybe just mentioning with PR changes and without this PR changes or something

GitHK:
> > > @GitHK are there any non-AI proofs that docker actually pulls already gzipped chunks?
> > 
> > 
> > @YuryHrytsuk As you asked I have ran some tests agains AWS master. This will fix the issue.
> > Notes on the screenshots below:
> > ```
> > * The mostly `red` line is the test without the headers on the registry
> > 
> > * The mostly `blue` line is the test with the suggested headers added to the registry
> > 
> > * The `red` test was shorter then the `blue` I can't tell by how much
> > 
> > * you can tell the test is ongoing because CPU goes up and network usage goes up (less evident since it always goes up a bit)
> > 
> > * in the `blue` test RAN stays flat no more pressure from the registry, no more restarts
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ...
> 
> Thanks for the test.
> 
> > The mostly blue line is the test with the suggested headers added to the registry
> 
> I am a bit confused by the wording. The PR suggests to remove headers (not to add them). Could you elaborate what is blue / red line? Maybe just mentioning with PR changes and without this PR changes or something

Don't get confused by my wording, I have applied the exact changes form this PR (modifies an existing header and adds a new one).

TL;DR:

- `red line`: contro test, same conditions and in prod
- `blue line`: test after changes applied. RAM usage no longer spikes all the rest remains the same (as expected) 👍

YuryHrytsuk:
Post factum: it actually helped. good job!

mrnicegyu11:
I read through this again as @sanderegg assigned me to the issue, and I think this fix is appropriate for now. i dont see a reason to unlimit or increase the CPU limit as of today, from those (indeed very nice!) investigations of @GitHK 

will close the issue https://github.com/ITISFoundation/osparc-simcore/issues/9013 as fixed for now

PR #1395, opened 2026-04-15
🐛 Fix cadvisor metrics dropped by excessive node labels on GPU nodes

mrnicegyu11:
## Problem

Since March 26, cadvisor container metrics (`container_memory_usage_bytes`, `container_cpu_usage_seconds_total`, etc.) are missing from VictoriaMetrics on the internal Kubernetes master cluster.

VictoriaMetrics logs show:
```
ignoring series with 100 labels for ...; either reduce the number of labels
for this metric or increase -maxLabelsPerTimeseries=40
```

## Root Cause

GPU nodes with the NVIDIA operator and Node Feature Discovery (NFD) have 60+ node labels (`feature.node.kubernetes.io/*`, `nvidia.com/*`). The default kubelet `VMNodeScrape` config has a `labelmap` rule that copies **all** `__meta_kubernetes_node_label_(.+)` onto every scraped metric. This pushes total labels past `maxLabelsPerTimeseries=40`, causing VMsingle to silently drop the entire series.

## Fix

Add `metricRelabelConfigs` to the kubelet vmScrape override that drops the noisy labels after scraping:
- `feature_node_kubernetes_io_.*` (NFD hardware/CPU/kernel feature labels)
- `nvidia_com_.*` (GPU model, memory, CUDA version, etc.)

These labels exist for **pod scheduling** (node selectors/affinities), not observability. Useful node identifiers (`node`, `instance`, `job`, `ops`, `worker`) are preserved. GPU info remains queryable via `kube_node_labels` from kube-state-metrics.

The 3 upstream default `metricRelabelConfigs` rules are also preserved (drop `uid`, `id|name`, `rest_client_*` metrics).

Fixes #1383


mrnicegyu11:
It will be better to exclude the labels at https://github.com/ITISFoundation/osparc-ops-environments/blob/main/charts/nvidia-gpu-operator/values.yaml.gotmpl

mrnicegyu11:
Added NFD labelWhitelist configuration to charts/nvidia-gpu-operator/values.yaml.gotmpl to exclude the labels at the source.

Configured node-feature-discovery.worker.config.core.labelWhitelist: "pci-10de\\." which restricts NFD to only create NVIDIA PCI vendor labels (needed for GPU node detection), eliminating the 60+ feature.node.kubernetes.io/* labels at the source rather than dropping them downstream in VictoriaMetrics.


**Update**: Corrected the NFD configuration approach after deeper research.

Instead of `core.labelWhiteList` (regex filter), using `core.labelSources: ["pci"]` which is the proper way to restrict which NFD feature sources create node labels.

**What this does:**
- By default, NFD enables ALL label sources (cpu, kernel, system, memory, network, storage, usb, pci), which creates 50+ `feature.node.kubernetes.io/*` labels on GPU nodes
- Setting `labelSources: ["pci"]` restricts NFD to only create PCI vendor-presence labels
- The PCI source is already configured upstream (by the GPU operator chart defaults) with `deviceClassWhitelist: ["02", "0200", "0207", "0300", "0302"]` and `deviceLabelFields: [vendor]`, so it only produces ~2-4 labels like `pci-10de.present`

**What this does NOT fix:**
- `nvidia.com/*` labels (e.g. `nvidia.com/gpu.product`, `nvidia.com/gpu.memory`) come from GFD (GPU Feature Discovery), which is a separate component from NFD. These are needed for GPU scheduling (node selectors by GPU type) and cannot be excluded at the source. They would still need `metricRelabelConfigs` in VictoriaMetrics to prevent metric label explosion.

mrnicegyu11:
Triple-checked the `core.labelSources: ["pci"]` approach against 3 independent sources:

1. **[NFD Worker Config Reference (v0.16)](https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/reference/worker-configuration-reference.html)** — confirms `core.labelSources` is the correct field (not deprecated `core.sources`), default is `[all]`, and it only affects label publishing (not feature detection for NodeFeatureRules).

2. **[NFD Helm chart defaults (v0.16.6)](https://github.com/kubernetes-sigs/node-feature-discovery/blob/v0.16.6/deployment/helm/node-feature-discovery/values.yaml)** — confirms `labelSources: [all]` is the default, explaining why 60+ labels appear without override.

3. **[NVIDIA GPU Operator docs (v26.3.1)](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html)** — confirms GPU nodes are identified solely by `feature.node.kubernetes.io/pci-10de.present=true`, so restricting to PCI source is safe.

`nvidia.com/*` labels (from GFD) still need metric-level filtering since they're a separate component needed for GPU scheduling.

mrnicegyu11:
@YuryHrytsuk re-review please, note this comment
image
sadly it seems that dropping labels in victoriametrics is still necessary

YuryHrytsuk:
@mrnicegyu11 I still do not understand, do we need `feature.node.kubernetes.io/*` labels? 

The fact that we need  `nvidia.com/*` labels is clear

mrnicegyu11:
Todo:
- check if the existing labels on the k8s nodes (machines) are removed!

YuryHrytsuk (approved):
talked in person.

pci feature labels  are needed but all others feature labels can be dropped.

Extra: node labels need to be manually pruned.

mrnicegyu11:
label cleanup on osparc-master-speag via

```
ubuntu@devman1:~$ for node in $(kubectl get nodes -o name); do
  labels=$(kubectl get "$node" -o json | jq -r '.metadata.labels | keys[] | select(test("^feature.node.kubernetes.io/(?!pci)"))' )
  if [ -n "$labels" ]; then
    patch=""
    for label in $labels; do
      patch="$patch $label-"
    done
    kubectl label "$node" $patch
  fi
done
node/devgpu1 unlabeled
node/devman1 unlabeled
node/devops1 unlabeled
node/devsim1 unlabeled
```

verification via
```
ubuntu@devman1:~$ kubectl get nodes -o json | jq -r '.items[] | .metadata.labels | to_entries[] | select(.key | test("^feature.node.kubernetes.io/(?!pci)"))'

```

returning nothing

mrnicegyu11:
related fix https://github.com/ITISFoundation/osparc-ops-environments/pull/1412

PR #1420, opened 2026-06-08
draft: scrape syslogs from kubernetes nodes

YuryHrytsuk:
## Checklist
- [ ] I tested and it works


YuryHrytsuk:
To be continued later. most likely we shall stick to existing approach (we use in docker swarm, on autoscaled node and comp. clusters when we use rsyslog

PR #1421, opened 2026-06-08
simcore invitations / datcore-adapter: raise cpu limit to fix failing healthcheck

YuryHrytsuk:
## What do these changes do?
These service fail to start on master since healthcheck execution takes more than healthcheck timeout. Services can never start. Raising CPU Limit to 0.5 together with changes from the related PR fix this problem

Using 0.5 CPU
image

## Related issue/s
* https://github.com/ITISFoundation/osparc-simcore/issues/9182

## Related PR/s
* https://github.com/ITISFoundation/osparc-simcore/pull/9211

## Checklist
- [x] I tested and it works

### DIFF
--- a/services/simcore/docker-compose.yml.j2
+++ b/services/simcore/docker-compose.yml.j2
@@ -224,7 +224,7 @@ services:
           cpus: "0.1"
           memory: "64M"
         limits:
-          cpus: "0.2"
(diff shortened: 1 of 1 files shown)


giancarloromeo (approved):
💸

PR #1425, opened 2026-06-15
Ensure otel collector port propagates to chatbot

bisgaard-itis:
## What do these changes do?
- Ensure OTEL collector port is propagated to otel collector endpoint in chatbot. The hardcoded value was wrong and because of that we could not see the chatbot traces in Grafana Tempo.
- I already tested this by changing the env vars in portainer.
## Related PR/s
- https://git.speag.com/oSparc/s4l-chatbox/-/merge_requests/132

## Checklist
- [x] I tested and it works


mrnicegyu11 (approved):
try it first if it works, but ok ;)

mrnicegyu11 (approved):

PR #1427, opened 2026-06-18
2026/fix/traefik redirect

mrnicegyu11:
## What do these changes do?
Always strip trailing slashes from ops-traefik-managed redirects.

This is needed to get osparc-io-anonymous-templates (like sarvalidation-site) to work, osparc doesnt accept trailing slashes.

This is already put into place in prod (manually in portainer)
## Checklist
- [x] I tested and it works


services/traefik/docker-compose.yml.j2
```diff
@@ -167,7 +167,7 @@ services:
         #     Use https://regex101.com/r/58sIgx/2 for regex explanation and experimentation.
         #     Below we include dollar escaping and j2 expressions. It is not clean / pure regex
         #     You can fetch baked and clean regex from traefik dashboards.
-        - traefik.http.middlewares.redirect-{{ from_domain_no_dots }}.redirectregex.regex=^https?://((?:[a-zA-Z0-9-]+\.)*)*{{ from_domain }}(.*)$$
+        - traefik.http.middlewares.redirect-{{ from_domain_no_dots }}.redirectregex.regex=^https?://((?:[a-zA-Z0-9-]+\.)*)*{{ from_domain }}(.*?)/?$$
```
YuryHrytsuk, line 170: what happens if `/` slash is needed? Current implementation keeps trailing slash only if it is present originally and gives the responsibility to link requester.

I feel like the right place to fix it is original link generation.

mrnicegyu11: yes, formally correct.

but, from a usability point of view, users might have their browsers, depending on how they type, which browser etc, go to https://sarvalidation.site/ or https://sarvalidation.site . 

Both should work, I guess, both are valid.

There is no impact on existing redirects (only sim4life-lite.io), but if we need more complex redirects in the future this might become an issue, in the future. If the trailing slash is ever needed. But since osparc anyway refuses a trailing slash completely and misbehaves, I think we are safe stripping it for now ;)

YuryHrytsuk: Well, if some web server does not accept `http://sarvalidation.site/` then we should not fix it in traefik. This is wrong as another web servers may require trailing slash and traefik will never be able to work with them.

Am I missing something? Who is requesting `http://sarvalidation.site/` and why is trailing `/` not working?

mrnicegyu11: users (you, me, mark, nik, someones mum) are free to type either verbatim `https://sarvalidation.site/` or `https://sarvalidation.site` in the browser. The browser to the best of my understanding is free to add/strip trailing slashes as well, since it is identical.

But, `https://sarvalidation.site` is by the traefik redirect transformed into `https://osparc.io/#/study/a66304c8-2909-599b-bfbe-eb857929d3d5`, and a redirect is done

And, `https://sarvalidation.site/` is - without this PR - transformed into `https://osparc.io/#/study/a66304c8-2909-599b-bfbe-eb857929d3d5/`. The latter is not a valid template URL for the osparc-simcore webserver. Why that is so, I dont know. But, to be iteroperable with osparc and have users (==clients) type both `https://sarvalidation.site` and `https://sarvalidation.site/`, we need to strip the trailing slash in the redirect 🤷‍♂️ 

The other option is to fix it in the osparc-simcore webserver, but a.) i kinda dont wanna do that and b.) at least today, to make the sarvalidation.site work, we need the trailing-slash-strip... if the osparc-simcore change is done, we could remove the trailing-slash stripping again

mrnicegyu11: FYI per the RFC 3986: equivalence of `http://example.com` and `http://example.com/` holds. For any non-root path, `https://example.com/path` and `https://example.com/path/` are not the same resource per RFC 3986

YuryHrytsuk (approved):
discussed in person.

it does change redirect behaviour by not preserving trailing slash when redirecting. This might be breaking things in case there is a path with trailing `/` which then redirected to path without trailing `/` and if web servers are sensitive to trailing slash

matusdrobuliak66 (approved):

mrnicegyu11:
https://github.com/ITISFoundation/osparc-simcore/pull/9284 would make this PR obsolete

mrnicegyu11:
Merged PR https://github.com/ITISFoundation/osparc-simcore/pull/9284 , this can be closed

PR #1442, opened 2026-07-06
Run renovate daily

YuryHrytsuk:
## What do these changes do?
Fix existing schedule which is most likely wrong. New value is taken from https://docs.renovatebot.com/key-concepts/scheduling/#recommended-cron-syntax

## Related PR/s
* value was originally introduced in https://github.com/ITISFoundation/osparc-ops-environments/pull/1437

## Checklist
- [ ] I tested and it works


mrnicegyu11:
ok

PR #1453, opened 2026-07-08
:recycle: add api worker to monitored network

mrnicegyu11:
## What do these changes do?
In order for the api-worker to talk to the chatbot-backend (usecase: requests from s4l-desktop via simcore api server), the api-worker and chatbot must be in the same docker network. 

Before: 
Chatbot: monitored, public
Api-Worker: simcore-default

After:
Chatbot: monitored, public
Api-Worker: simcore-default, monitored

@bisgaard-itis and me went through all options and decided this is not  the cleanest, but the most pragmatic option. We are both excited for :wheel: kubernetes to bring proper networking segementation options.

## Checklist
- [ ] I tested and it works


bisgaard-itis (approved):
Thanks a lot for the quick fix 👍🏻

PR #1457, opened 2026-07-14
charts Makefile: verify helmfile version

YuryHrytsuk:
## What do these changes do?

Add a check in makefile verifying installed helmfile version is up-to-date. Print a warning otherwise. 

Add comments and notes about how to choose helmfile version and to keep it in sync with other places

## Checklist
- [x] I tested and it works

### DIFF
--- a/charts/Makefile
+++ b/charts/Makefile
@@ -38,7 +38,7 @@ CHART_TEMPLATES := $(wildcard $(CHART_TEMPLATE_DIR)/*)
 	fi
 	@if ! helmfile --version | grep -q "$(HELMFILE_EXPECTED_VERSION)"; then \
 		echo "WARNING: installed helmfile version does not match"> /dev/stderr; \
-		echo "Expected helmfile version:'$(HELMFILE_EXPECTED_VERSION)'" > /dev/stderr; \
+		echo "Expected helmfile version:'$(HELMFILE_EXPECTED_VERSION)'. Please, update!" > /dev/stderr; \
 		sleep 3s; \
 	fi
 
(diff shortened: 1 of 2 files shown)


charts/Makefile
```diff
@@ -23,15 +30,20 @@ CHART_TEMPLATES := $(wildcard $(CHART_TEMPLATE_DIR)/*)
...
+.verify-helmfile-installation:
 	@if ! command -v helmfile >/dev/null 2>&1; then \
-			echo "'helmfile' is not installed. Install it to continue ...";\
+		echo "'helmfile' is not installed. Install version $(HELMFILE_EXPECTED_VERSION) to continue ..."; \
+		exit 1; \
+	fi
+	@if ! helmfile --version | grep -q "$(HELMFILE_EXPECTED_VERSION)"; then \
+		echo "WARNING: installed helmfile version does not match"> /dev/stderr; \
+		echo "Expected helmfile version:'$(HELMFILE_EXPECTED_VERSION)'" > /dev/stderr; \
+		sleep 3s; \
```
sanderegg, line 42: why the sleep?

sanderegg (approved):

charts/Makefile
YuryHrytsuk: so it does not go unnoticed by output of helmfile CLI

... and it creates incentive to keep it up-to-date :upside_down_face:

sanderegg: you could propose to update then.

YuryHrytsuk: what do you mean? `echo "Please update"`?

sanderegg: `Do you wish to update? [Y/n]`

YuryHrytsuk: I will add a string `Please, update!`. But I want incorporate installation of the tool in Makefile.

I want to keep Makefile clean, short and robust. Adding installation of helmfile would make it already longer and also it would be brittle. For now, I would like to stay away from it and keep it simple and robust

sanderegg: ok... np, that was mostly because you asked.

PR #1458, opened 2026-07-14
Kubernetes CI: use real deployment structure

YuryHrytsuk:
## What do these changes do?
**Changes**
* Use the same folder structure for CI as for real deployments.
* Add `trivy-scan` target to makefile
* Add helmfile static checks (`lint`, `template`) to CI Tests

**Reasoning**
With this approach we can re-use existing logic (e.g. Makefiles) as is without any special handling. Using makefile as is in CI is handy. 

For example, now one can run CI tests manually by pointing `.config.location` to `./charts/.ci-deployment` and running `make -C charts trivy-scan` (exactly what CI does)

## Checklist
- [x] I tested and it works


.github/actions/setup-helmfile/action.yml
```diff
@@ -0,0 +1,16 @@
...
+  using: composite
+  steps:
+    - name: Configure config location
+      shell: bash
+      run: echo $(realpath charts/.ci-deployment/repo.config) > .config.location
+
+    - name: Install helmfile
+      uses: helmfile/helmfile-action@v2.4.4
+      with:
+        helmfile-version: v0.165.0 # keep in sync with charts/Makefile and version in running clusters
```
sanderegg, line 14: would a test be needed for this?

sanderegg (approved):
thanks

.github/actions/setup-helmfile/action.yml
YuryHrytsuk: the problem is I don't know a test that would be nice and robust for this. Do you have ideas?

YuryHrytsuk: Probably, the test would look like a hack just by nature of a problem it should solve

sanderegg: not sure what you mean by that. but currently I guess the other end of the part that uses that version does not know about this. so how do you mean to keep them in sync?

YuryHrytsuk: manually. just by reading the comment next to the version

YuryHrytsuk: discussed in person. this does not block current PR since it relates to a problem outside of scope of this PR

PR #1465, opened 2026-07-19
pre-commit: improve `helm-dependency-update` hook

YuryHrytsuk:
## What do these changes do?
Only run `helm dependence update` when `Chart.yaml` file changed

## Checklist
- [x] I tested and it works


mrnicegyu11 (approved):
nice thx

PR #1473, opened 2026-07-30
Rabbit Cluster: handle network partitions (split brain)

YuryHrytsuk:
## What do these changes do?
On on-premise deployments we sometimes get rabbit cluster in split brain state. At some point a particular node cannot reach the other 2 and we end up running (so-to-speak) 2 Rabbit Clusters, since all nodes are running and HA-Proxy (LB) routing to all of them. This breaks clients as they keep connecting to different clusters.

Solution: use cluster partition handling strategy `pause_minority`. In pause-minority mode RabbitMQ will automatically pause cluster nodes which determine themselves to be in a minority. This shall solve the issue

Further read:
* https://www.rabbitmq.com/docs/4.1/partitions#automatic-handling
* https://www.rabbitmq.com/docs/4.1/partitions#pause-minority

Disclaimer:
* we need to keep rabbit cluster running and see how it behaves with a new strategy. I expect cluster to be healthy and coherent when network partition occurs, the 2 nodes that can see each other shall keep running the cluster while 3rd node shall be stopped. e2e-test will later spot this as node3 tests shall fail

## Checklist
- [ ] I tested and it works

### DIFF
--- a/services/rabbit/Makefile
+++ b/services/rabbit/Makefile
@@ -7,6 +7,7 @@ include ${REPO_BASE_DIR}/scripts/common.Makefile
 # Targets
 #
 
+# All cluster operations (start, stop, etc) are included below
 include ${REPO_BASE_DIR}/services/rabbit/.operations.Makefile
 
 #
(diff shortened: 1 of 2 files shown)


mrnicegyu11 (approved):
wow we need this, thanks a lot!

PR #1480, opened 2026-08-03
Kubernetes: update node selectors :warning:

YuryHrytsuk:
## What do these changes do?
Follow new label format. See changes in the related PR

Manual action required:
* Delete old node labels `kubectl label node --all ops-`

## Related PR/s
* https://git.speag.com/oSparc/osparc-ops-deployment-configuration/-/merge_requests/2208

## Checklist
- [x] I tested and it works

### DIFF
--- a/charts/cert-manager/values.common.yaml.gotmpl
+++ b/charts/cert-manager/values.common.yaml.gotmpl
@@ -6,7 +6,7 @@ cert-manager:
     keep: true
 
   nodeSelector:
-    ops: "true"
+    node-role.kubernetes.io/ops: ""
 
   # controller
   replicaCount: 1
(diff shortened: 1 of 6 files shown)


charts/cert-manager/values.common.yaml.gotmpl
```diff
@@ -6,7 +6,7 @@ cert-manager:
     keep: true
 
   nodeSelector:
-    ops: "true"
+    node-role.kubernetes.io/ops: ""
```
sanderegg, line 9: has this something to do with 
`"io.simcore.instance-role":"autoscaled-dynamic"`?

if yes, we should use the same practices

sanderegg (commented):
as good as it gets. but there is a question

charts/cert-manager/values.common.yaml.gotmpl
sanderegg: very good and very exciting!

sanderegg (approved):
perfect

